PR #2, opened 2018-12-08
improvements to IME support

bfredl:
- set the client window of the im_context (needed for some events)
- esapce any "<" in commited input, not only a single one
- make external preedit track cursor position

NB: I made the `Grid` mutable during initialization, maybe this could be expressed in some better way that doesn't need it.

### DIFF
--- a/src/ui/grid/grid.rs
+++ b/src/ui/grid/grid.rs
@@ -63,6 +63,8 @@ pub struct Grid {
     /// Pointer position for dragging if we should call callback from
     /// `connect_motion_events_for_drag`.
     drag_position: Arc<ThreadGuard<(u64, u64)>>,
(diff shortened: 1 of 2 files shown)


vhakulinen:
Can you give me an example of what kind of input issue this fixes?

bfredl:
Basically an IME can forward input in two ways: it can either commit a unicode string, or it can send syntethic keypresses. If the IME commits the text `<<`, then both `<` needs to be escaped to `<lt>`. In order for synthetic keypresses to work, the input context must be bound to a gdk surface (this will also affect preedit position, so it should be changed to the active drawing area, if there is one per grid). As an example of a synthetic keypress, my IME translates space+alphnum to altgr+alphanum, but just space is still a space.

PR #13, opened 2019-03-02
Fix issue with packaging

GabMus:
See *trevdev*'s comment on the [AUR page for gnvim-git](https://aur.archlinux.org/packages/gnvim-git/) for details


vhakulinen:
Thanks!

PR #15, opened 2019-03-02
Fix sync calls

### LINKED ISSUE #10: Synchronous commands after ui_attach
I've run into a couple of related issues. The problem seems to be that if synchronous commands are used after ui_attach, there's a possibility they will deadlock if nvim needs to get input from the user before proceeding (I know very little about the RPC interface). The cases I've run into are if there's an error in .vimrc, or if one of the files being edited has a swap file. The synchronous commands are:
- `nvim.command` for opening the files as tabs. Can these be appended to the nvim command line instead, with the usual `-p` option for tabs?
- `nvim.ui_try_resize`

Commenting out the code for these allows gnvim to proceed.

vhakulinen:
Closes #10

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -60,7 +60,7 @@ struct Options {
     )]
     gnvim_rtp: String,
 
-    /// Files to open. Files after the first one are opened in new tabs.
+    /// Files to open.
     #[structopt(value_name = "FILES")]
     open_files: Vec<String>,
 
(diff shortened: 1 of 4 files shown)


bfredl:
This works, as long as the actual size is larger than the initial guessed size `80, 30`. Startup errors will be displayed close to line 30 though, rather than the actual bottom of the screen.

vhakulinen:
Yea, complete fix requires bigger change.

bfredl:
Alternatively (on top of this change), implementing ext_messages would side-step this initial size issue.

PR #30, opened 2019-04-24
Properly set cursor position in cmdline input

### LINKED ISSUE #24: Panic when opening Netrw
### How to reproduce

- `mv .config/nvim/init.vim .config/nvim/init.vim_` (optional, bring the config back afterward)
- Open gnvim with the terminal
- Type `:e .`

### What happens

The program panics leaving the output:

```
thread 'main' panicked at 'byte index 2 is out of bounds of ``', src/libcore/str/mod.rs:2085:9
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

### What is expected

Vim's netrw is opened.

vhakulinen:
Fixes #24

### DIFF
--- a/src/ui/cmdline.rs
+++ b/src/ui/cmdline.rs
@@ -286,13 +286,11 @@ impl CmdlineInput {
             buffer.insert_markup(&mut iter, &markup);
         }
 
-        self.current_level = content.level;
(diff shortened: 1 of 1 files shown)


badosu:
I confirm I can't reproduce the earlier bug. It seems it depends on the occasions, will try again later

PR #37, opened 2019-04-30
Add Ignore variant to RedrawEvent

agraven:
gnvim prints a lot of messages to the console about unknown events which are valid, but not handled. This adds a no-op `Ignore` variant to `RedrawEvent` to silence those messages.

Storing the event string could be useful for debugging purposes, but it's currently unused, so it's worth considering whether it should be kept


vhakulinen:
Thanks!

PR #42, opened 2019-05-09
Handle guicursor blinkon property

### LINKED ISSUE #17: How can I make the cursor NOT blink?
#### checkhealth

```
health#deoplete#check
========================================================================
## deoplete.nvim
  - OK: exists("v:t_list") was successful
  - OK: has("timers") was successful
  - OK: has("python3") was successful
  - OK: Python3.6.1+ was successful
  - INFO: If you're still having problems, try the following commands:
    $ export NVIM_PYTHON_LOG_FILE=/tmp/log
    $ export NVIM_PYTHON_LOG_LEVEL=DEBUG
    $ nvim
    $ cat /tmp/log_{PID}
    and then create an issue on github

health#nvim#check
========================================================================
## Configuration
  - OK: no issues found

## Performance
  - OK: Build type: RelWithDebInfo
    LuaJIT

## Remote Plugins
  - OK: Up to date

## terminal
  - INFO: key_backspace (kbs) terminfo entry: key_backspace=\177
  - INFO: key_dc (kdch1) terminfo entry: key_dc=\E[3~
  - INFO: $VTE_VERSION='4601'
  - INFO: $COLORTERM='truecolor'

## tmux
  - OK: escape-time: 0ms
  - INFO: $TERM: tmux-256color

health#provider#check
========================================================================
## Clipboard (optional)
  - OK: Clipboard tool found: xclip

## Python 2 provider (optional)
  - INFO: pyenv: Path: /home/dave/.pyenv/libexec/pyenv
  - INFO: pyenv: $PYENV_ROOT is not set. Infer from `pyenv root`.
  - INFO: pyenv: Root: /home/dave/.pyenv
  - INFO: Using: g:python_host_prog = "/usr/bin/python2.7"
  - INFO: Executable: /usr/bin/python2.7
  - INFO: Python version: 2.7.13
  - INFO: pynvim version: 0.3.2
  - OK: Latest pynvim is installed.

## Python 3 provider (optional)
  - INFO: pyenv: Path: /home/dave/.pyenv/libexec/pyenv
  - INFO: pyenv: $PYENV_ROOT is not set. Infer from `pyenv root`.
  - INFO: pyenv: Root: /home/dave/.pyenv
  - INFO: `g:python3_host_prog` is not set.  Searching for python3 in the environment.
  - INFO: Executable: /home/dave/.pyenv/versions/3.7.2/bin/python3
  - INFO: Python version: 3.7.2
  - INFO: pynvim version: 0.3.2
  - OK: Latest pynvim is installed.

## Ruby provider (optional)
  - INFO: Ruby: ruby 2.3.3p222 (2016-11-21) [x86_64-linux-gnu]
  - WARNING: `neovim-ruby-host` not found.
    - ADVICE:
      - Run `gem install neovim` to ensure the neovim RubyGem is installed.
      - Run `gem environment` to ensure the gem bin directory is in $PATH.
      - If you are using rvm/rbenv/chruby, try "rehashing".
      - See :help |g:ruby_host_prog| for non-standard gem installations.

## Node.js provider (optional)
  - WARNING: `node` and `npm` (or `yarn`) must be in $PATH.
    - ADVICE:
      - Install Node.js and verify that `node` and `npm` (or `yarn`) commands work.

```

badosu:
This is an initial effort to support setting blinking on gnvim.

For retrocompatibility, a runtime setting is added indentical to the previous effect.

To disable the blinking `set guicursor+=a:blinkon0`.

Fixes #17

### DIFF
--- a/src/ui/grid/grid.rs
+++ b/src/ui/grid/grid.rs
@@ -402,9 +402,15 @@ impl Grid {
         let mut ctx = self.context.borrow_mut();
         let ctx = ctx.as_mut().unwrap();
 
-        ctx.cursor_alpha += 0.05;
-        if ctx.cursor_alpha > 2.0 {
-            ctx.cursor_alpha = 0.0;
+        if ctx.cursor_blink_on > 0 {
+            // Assuming a 60hz framerate
+            ctx.cursor_alpha += 100.0 / (6.0 * ctx.cursor_blink_on as f64);
+
+            if ctx.cursor_alpha > 2.0 {
+                ctx.cursor_alpha = 0.0;
+            }
+        } else {
+            ctx.cursor_alpha = 0.5;
         }
 
         let (x, y, w, h) = {
@@ -445,6 +451,7 @@ impl Grid {
         let mut ctx = self.context.borrow_mut();
         let ctx = ctx.as_mut().unwrap();
 
+        ctx.cursor_blink_on = mode.blink_on;
         ctx.cursor_cell_percentage = mode.cell_percentage;
     }
 
--- a/src/nvim_bridge.rs
+++ b/src/nvim_bridge.rs
@@ -157,6 +157,8 @@ impl Default for CursorShape {
 
 #[derive(Default, Clone)]
 pub struct ModeInfo {
+    /// The cursor blinking period (in ms)
+    pub blink_on: u64,
     pub cursor_shape: CursorShape,
     /// The cursor's width (in percentages, from 0..1).
     pub cell_percentage: f64,
@@ -166,6 +168,9 @@ pub struct ModeInfo {
 impl ModeInfo {
     fn set(&mut self, prop: &str, val: Value) {
         match prop {
+            "blinkon" => {
+                self.blink_on = unwrap_u64!(val);
+            }
             "cursor_shape" => {
                 self.cursor_shape = CursorShape::from_string(unwrap_str!(val))
             }
--- a/src/ui/grid/context.rs
+++ b/src/ui/grid/context.rs
@@ -28,7 +28,9 @@ pub struct Context {
     pub cursor: (u64, u64),
     /// Cursor alpha color. Used to make the cursor blink.
     pub cursor_alpha: f64,
-    /// Width of the curosr.
+    /// The duration of the cursor blink
+    pub cursor_blink_on: u64,
+    /// Width of the cursor.
     pub cursor_cell_percentage: f64,
     /// Color of the cursor.
     pub cursor_color: Color,
@@ -75,6 +77,7 @@ impl Context {
 
             cursor: (0, 0),
             cursor_alpha: 1.0,
+            cursor_blink_on: 0,
             cursor_cell_percentage: 1.0,
             cursor_color: Color::from_u64(0),
             busy: false,


runtime/plugin/gnvim.vim
```diff
@@ -4,6 +4,9 @@ endif
 
 let g:gnvim_runtime_loaded = 1
 
+" Set blinking cursor as default
+set guicursor+=a:blinkon333
```
vhakulinen: Gnvim should modify `guicursor` (or other options for that manner). Instead, we should provide documentation to the users which instructs them on this topic.

src/ui/grid/grid.rs
```diff
@@ -402,9 +402,15 @@ impl Grid {
...
-        if ctx.cursor_alpha > 2.0 {
-            ctx.cursor_alpha = 0.0;
+        if ctx.cursor_blink_on > 0 {
+            // Assuming a 60hz framerate
+            ctx.cursor_alpha += 100.0/(6.0 * ctx.cursor_blink_on as f64);
+
+            if ctx.cursor_alpha > 2.0 {
+                ctx.cursor_alpha = 0.0;
+            }
+        else {
```
vhakulinen: You're missing closing `}` here.

vhakulinen:
This is looking good. One major thing to consider is that instead of setting the cursor opacity to 50% when the blinking is off, I think gnvim should do the same as the TUI does. Which is to invert the colors on the cursor column (e.g. foreground color becomes the background color, and background color comes foreground color).

runtime/plugin/gnvim.vim
badosu: That's fine by me, the reason I included this was to make it compatible with the previous implementation (blink on by default).

src/ui/grid/grid.rs
badosu: What do you mean? Isn't it closed on line 414?

vhakulinen:
Can you run `cargo fmt` before merge? We can figure out the color inverting later.

badosu:
@vhakulinen Done.

By the way, I was trying some approaches for the reverse colors blink.

If you don't mind sharing, how would you approach this? I've tried understanding how to split segments and passing a custom hldef here: https://github.com/badosu/gnvim/blob/9789ff84c4597fcc84625aaa27e084c63c846d47/src/ui/grid/render.rs#L13-L22.

vhakulinen:
> If you don't mind sharing, how would you approach this?

Sure! My goto place to add the logic for inverting colors under cursor would go in [cursor_goto](https://github.com/vhakulinen/gnvim/blob/356f79dd257c30ae24d6c6d61460b586045b70a5/src/ui/grid/grid.rs#L310). (At least) every time `cursor_goto` us called, we need to get the cell under the cursor - currently we can get the `Leaf`, but we are only interested in the individual cell so the `Row` API needs to be extended[1]. Once we have that, we need to do pretty much the same thing we do in [put_segments](https://github.com/vhakulinen/gnvim/blob/356f79dd257c30ae24d6c6d61460b586045b70a5/src/ui/grid/render.rs#L27-L121) - just remember to flip the `hl.reverse`. I would try to keep the rendering logic combined between these two (grid rendering and the cursor logic), but not sure how feasible that is.

One thing to note in that approach is that the code paths for blinking cursor and non blinking cursor are extremely different. With blinking cursor, we (currently) leave the actual blink logic for the `draw` signal of the drawing area (with some help of `tick` method which it self is currently triggered by a bit of a lazy hack). With non blinking cursor the logic in the `draw` signal handler can (or at least should) be ditched and `cursor_goto` call should contain logic to draw the inverted cell to the drawing area's context.

1: I'm starting to think more and more that the rope structure behind the grid is way too much, but at least it makes the rendering easy. If there are compelling reasons to switch it to something else, I'm open for the idea.

PR #45, opened 2019-05-16
Improve .desktop file

elihunter173:
I added more detail to the `gnvim.desktop` file, most notably the MIME type and keywords, so that GNvim can be discovered as a text editor in specific "Select Application for ____ file type" dialogs.

I also added a description of the program in many different languages.

This was achieved by copying the `.desktop` file from Neovim and then modifying it to be appropriate for GNvim.


vhakulinen:
Thanks!

Issue #47, opened 2019-05-31
All popup menu icons are question mark icons


TypeAndProgram:
From the screenshots of GNvim in the wiki I assume that there should be icons of boxes, < >, etc. in the popup menu for auto completions. However, as the title states, mine only shows circled question marks:

![image](https://user-images.githubusercontent.com/46855713/58734487-2edffe80-83bd-11e9-8f26-e5a4d3422f60.png)

Am I missing a package/font for my system? The GNvim runtime is loaded, and I have not installed GNvim; I am running it w/out installing using the `GNVIM_RUNTIME_PATH=/path/to/gnvim/runtime cargo run` command, as an FYI (although IIRC it did the same thing when it was installed on my previous Linux install).

badosu:
This is the same for me too, I wonder if you're using a LSP plugin that is not [gnvim-lsp](https://github.com/vhakulinen/gnvim-lsp)?

I use [coc.nvim](https://github.com/neoclide/coc.nvim) and this is how it looks like:

![Screenshot from 2019-06-03 23-08-03](https://user-images.githubusercontent.com/347552/58846115-78964680-8654-11e9-9828-0036cec4d99b.png)

TypeAndProgram:
No, I use [LanguageClient-Neovim](https://github.com/autozimu/LanguageClient-neovim). Is gnvim-lsp required to have the icons? If so, do you think that info could be added to the wiki? I’ll try switching to vim-lsp with gnvim-lsp and let you know what happens.

TypeAndProgram:
Ok, so the problem is the language server. Using [vim-lsp](https://github.com/prabirshrestha/vim-lsp) (with or without [gnvim-lsp](https://github.com/vhakulinen/gnvim-lsp)), the icons show correctly:

![image](https://user-images.githubusercontent.com/46855713/58879355-20366780-869b-11e9-8a39-de728c2f6dbe.png)

Is there is a way to add this functionality to gnvim for other language clients like [LanguageClient-Neovim](https://github.com/autozimu/LanguageClient-neovim) as well? If not or until this is added, can this info be added to the wiki/README?

Also, could it not be set for gnvim to default to no icons in the popup menu if vim-lsp is absent?

badosu:
> Is there is a way to add this functionality to gnvim for other language clients like LanguageClient-Neovim as well? If not or until this is added, can this info be added to the wiki/README?

[...]

PR #49, opened 2019-06-04
Support other LSPs completion icons

badosu:
This is an initial effort to show how support for other LSP plugins may work.

Screenshot from 2019-06-04 14-57-50

This implementation works, but would gladly appreciate feedback if a better approach is possible.

See also:

- https://github.com/vhakulinen/gnvim/issues/47#issuecomment-498773230
- https://github.com/neovim/neovim/issues/10119

### DIFF
--- a/src/nvim_bridge.rs
+++ b/src/nvim_bridge.rs
@@ -242,33 +242,43 @@ pub enum CompletionItemKind {
 }
 
 impl From<&str> for CompletionItemKind {
+    // Returns CompletionItemKind from a string
+    //
+    // Lower case kinds are vim-lsp
+    //   https://github.com/prabirshrestha/vim-lsp/blob/2b583fefa20b7b1a5e7481a93fb6f1fee67e0846/autoload/lsp/omni.vim#L4-L28
+    // Single characters are coc.nvim
+    //   https://github.com/neoclide/coc.nvim/blob/909710fddb04d383e5546b0f869c44f395a80d02/src/languages.ts#L143-L167
+    // Pascal cased kinds are LanguageClient-neovim
+    //   https://github.com/autozimu/LanguageClient-neovim/blob/0ac444affdff8db699684aa4cf04c2cb0daf0286/rplugin/python3/denite/lsp/protocol.py#L48-L55
     fn from(from: &str) -> Self {
         match from {
-            "class" => CompletionItemKind::Class,
+            "class" | "C" | "Class" => CompletionItemKind::Class,
             "color" => CompletionItemKind::Color,
-            "constant" => CompletionItemKind::Constant,
-            "constructor" => CompletionItemKind::Constructor,
-            "enum" => CompletionItemKind::Enum,
-            "enum member" => CompletionItemKind::EnumMember,
-            "event" => CompletionItemKind::Event,
-            "function" => CompletionItemKind::Function,
-            "file" => CompletionItemKind::File,
-            "folder" => CompletionItemKind::Folder,
-            "field" => CompletionItemKind::Field,
-            "interface" => CompletionItemKind::Interface,
-            "keyword" => CompletionItemKind::Keyword,
-            "method" => CompletionItemKind::Method,
-            "module" => CompletionItemKind::Module,
-            "operator" => CompletionItemKind::Operator,
-            "property" => CompletionItemKind::Property,
-            "reference" => CompletionItemKind::Reference,
+            "constant" | "Constant" => CompletionItemKind::Constant,
+            "constructor" | "Constructor" => CompletionItemKind::Constructor,
+            "enum" | "Enum" => CompletionItemKind::Enum,
+            "enum member" | "Enum Member" => CompletionItemKind::EnumMember,
+            "event" | "E" | "Event" => CompletionItemKind::Event,
+            "file" | "F" | "File" => CompletionItemKind::File,
+            "field" | "m" | "Field" => CompletionItemKind::Field,
+            "folder" | "Folder" => CompletionItemKind::Folder,
+            "function" | "Function" => CompletionItemKind::Function,
+            "interface" | "I" | "Interface" => CompletionItemKind::Interface,
+            "keyword" | "k" | "Key" => CompletionItemKind::Keyword,
+            "method" | "f" | "Method" => CompletionItemKind::Method,
+            "module" | "M" | "Module" => CompletionItemKind::Module,
+            "operator" | "O" | "Operator" => CompletionItemKind::Operator,
+            "property" | "Property" => CompletionItemKind::Property,
+            "reference" | "r" => CompletionItemKind::Reference,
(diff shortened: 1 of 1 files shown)


vhakulinen:
I think this is good enough solution. I would like to see the missing ones added too (folder and file for example).

badosu:
@vhakulinen What do you mean, the coc ones? Coc.nvim has less granularity than vim-lsp so many icons are not mapped.

I used the ones I thought made sense, if you want to make a suggestion see https://github.com/neoclide/coc.nvim/blob/909710fddb04d383e5546b0f869c44f395a80d02/src/languages.ts#L143-L167 and let me know which icon you think is mapped better.

vhakulinen:
Hmh, lets get back to this once #50 is merged.

badosu:
@vhakulinen @smolck rebased.

I did not have found the time to test languageclient-neovim, will do that when I can. Meanwhile I hope we can merge this.

smolck:
@badosu I was able to test this with LanguageClient-Neovim, and it appeared to work and look as intended. Here is the diff:
```diff
diff --git a/src/nvim_bridge.rs b/src/nvim_bridge.rs
index be8e98a..6572334 100644
--- a/src/nvim_bridge.rs
+++ b/src/nvim_bridge.rs
@@ -244,31 +244,31 @@ pub enum CompletionItemKind {
 impl From<&str> for CompletionItemKind {
     fn from(from: &str) -> Self {
         match from {
-            "class" | "C" => CompletionItemKind::Class,
+            "class" | "C" | "Class" => CompletionItemKind::Class,
             "color" => CompletionItemKind::Color,
-            "constant" => CompletionItemKind::Constant,
-            "constructor" => CompletionItemKind::Constructor,
-            "enum" => CompletionItemKind::Enum,
-            "enum member" => CompletionItemKind::EnumMember,
-            "event" | "E" => CompletionItemKind::Event,
-            "function" => CompletionItemKind::Function,
-            "file" | "F" => CompletionItemKind::File,
-            "folder" => CompletionItemKind::Folder,
+            "constant" | "Constant" => CompletionItemKind::Constant,
+            "constructor" | "Constructor" => CompletionItemKind::Constructor,
+            "enum" | "Enum"=> CompletionItemKind::Enum,
+            "enum member" | "Enum Member" => CompletionItemKind::EnumMember,
+            "event" | "E" | "Event"=> CompletionItemKind::Event,
+            "function" | "Function" => CompletionItemKind::Function,
+            "file" | "F" | "File" => CompletionItemKind::File,
+            "folder" | "Folder" => CompletionItemKind::Folder,
             "field" | "m" => CompletionItemKind::Field,
-            "interface" | "I" => CompletionItemKind::Interface,
-            "keyword" | "k" => CompletionItemKind::Keyword,
-            "method" | "f" => CompletionItemKind::Method,
-            "module" | "M" => CompletionItemKind::Module,
-            "operator" | "O" => CompletionItemKind::Operator,
-            "property" => CompletionItemKind::Property,
+            "interface" | "I" | "Interface" => CompletionItemKind::Interface,
+            "keyword" | "k" | "Key" => CompletionItemKind::Keyword,
+            "method" | "f" | "Method" => CompletionItemKind::Method,
+            "module" | "M" | "Module" => CompletionItemKind::Module,
+            "operator" | "O" | "Operator" => CompletionItemKind::Operator,
+            "property" | "Property" => CompletionItemKind::Property,
             "reference" | "r" => CompletionItemKind::Reference,
             "snippet" => CompletionItemKind::Snippet,
-            "struct" | "S" => CompletionItemKind::Struct,
+            "struct" | "S" | "Struct" => CompletionItemKind::Struct,
             "text" => CompletionItemKind::Text,
-            "type parameter" | "T" => CompletionItemKind::TypeParameter,
+            "type parameter" | "T" | "Type Parameter" => CompletionItemKind::TypeParameter,
             "unit" | "U" => CompletionItemKind::Unit,
             "value" => CompletionItemKind::Value,
-            "variable" | "v" => CompletionItemKind::Variable,
+            "variable" | "v" | "Variable" => CompletionItemKind::Variable,
             _ => CompletionItemKind::Unknown,
         }
     }
```
Considering that this is a small addition, I think we could probably just include the support for LanguageClient-Neovim in this PR. What do you think?

badosu:
No problem, I'll include it, thanks!

badosu:
@vhakulinen @smolck I updated to use LanguaceClient-neovim icons as well.

Included some missing kinds from vim-lsp and took the liberty to merge some kinds icons.

Let me know your thoughts.

badosu:
Just pushed a separate commit for ease-of-review and reversibility extracting `CompletionItemKind` from `nvim_bridge`.

vhakulinen:
Interesting that LanguageClient-neovim doesn't comply to the lsp documentation: https://microsoft.github.io/language-server-protocol/specification#textDocument_completion. This PR should not add any new keywords, since the ones documented in the lsp spec are already there.

src/ui/popupmenu/completion_item_kind.rs
```diff
@@ -0,0 +1,105 @@
+#[derive(Clone)]
+pub enum CompletionItemKind {
```
vhakulinen: Keep this in the nvim_bridge file.

badosu: Done

badosu:
> This PR should not add any new keywords, since the ones documented in the lsp spec are already there.

You mean, we should restrict only to kinds specified by the documentation?

```
namespace CompletionItemKind {
	export const Text = 1;
	export const Method = 2;
	export const Function = 3;
	export const Constructor = 4;
	export const Field = 5;
	export const Variable = 6;
	export const Class = 7;
	export const Interface = 8;
	export const Module = 9;
	export const Property = 10;
	export const Unit = 11;
	export const Value = 12;
	export const Enum = 13;
	export const Keyword = 14;
	export const Snippet = 15;
	export const Color = 16;
	export const File = 17;
	export const Reference = 18;
	export const Folder = 19;
	export const EnumMember = 20;
	export const Constant = 21;
	export const Struct = 22;
	export const Event = 23;
	export const Operator = 24;
	export const TypeParameter = 25;
}
```

vhakulinen:
Yes. Otherwise we'll end up supporting unmaintainable set of completion kinds. We could also write our own documented set of supported kinds or even create a way to set the map from kind value to an icon through viml. But for now, lets stick with the kinds documented in lsp spec.

badosu:
@vhakulinen It seems this is a WIP at the moment, but some clients already have it implemented, see: https://github.com/microsoft/language-server-protocol/issues/343

Anyway, I reverted back to only standard kinds.

badosu:
Let me know if there are any issues pending.

smolck:
LGTM. Will definitely be nice to support the main LSPs 👍

vhakulinen:
Merged!

PR #50, opened 2019-06-04
Remove 'help-circle.svg' (question mark) icon from unknown autocompletions

smolck:
This is an idea of how to remove the 'help-circle.svg' icon from autocompletions that are unknown to clean up the look for completions with unsupported or unimplemented kinds (like those of [LanguageClient-Neovim](https://github.com/autozimu/LanguageClient-neovim)):

image

Any feedback would be much appreciated! Currently this works by filling in an empty image in the popupmenu if there is an error with the icon's `gdk_pixbuf` (which is created from returning the "None" string in the `get_icon_name_for_kind` function).

See #47.

### DIFF
--- a/src/nvim_bridge.rs
+++ b/src/nvim_bridge.rs
@@ -241,9 +241,9 @@ pub enum CompletionItemKind {
     Variable,
 }
 
-impl CompletionItemKind {
-    pub fn new_from_str(kind: &str) -> Self {
-        match kind {
+impl From<&String> for CompletionItemKind {
+    fn from(from: &String) -> Self {
+        match from.as_str() {
             "class" => CompletionItemKind::Class,
             "color" => CompletionItemKind::Color,
             "constant" => CompletionItemKind::Constant,
@@ -272,7 +272,10 @@ impl CompletionItemKind {
             _ => CompletionItemKind::Unknown,
         }
     }
-   pub fn is_unknown(&self) -> bool {
+}
+
+impl CompletionItemKind {
+    pub fn is_unknown(&self) -> bool {
         match self {
             CompletionItemKind::Unknown => true,
             _ => false,
@@ -284,6 +287,7 @@ impl CompletionItemKind {
 pub struct CompletionItem {
     pub word: String,
     pub kind: CompletionItemKind,
+    pub kind_raw: String,
     pub menu: String,
     pub info: String,
 }
@@ -748,7 +752,10 @@ fn parse_redraw_event(args: Vec<Value>) -> Vec<RedrawEvent> {
                     for item in unwrap_array!(args[0]) {
                         let item = unwrap_array!(item);
                         let word = unwrap_str!(item[0]).to_owned();
-                        let kind = CompletionItemKind::new_from_str(&unwrap_str!(item[1]).to_owned());
+                        let kind = CompletionItemKind::from(
+                            &unwrap_str!(item[1]).to_owned(),
+                        );
+                        let kind_raw = unwrap_str!(item[1]).to_owned();
                         let menu = unwrap_str!(item[2]).to_owned();
                         let info = unwrap_str!(item[3]).to_owned();
 
@@ -757,6 +764,7 @@ fn parse_redraw_event(args: Vec<Value>) -> Vec<RedrawEvent> {
                             kind,
                             menu,
                             info,
+                            kind_raw,
                         });
                     }
 
--- a/src/nvim_bridge.rs
+++ b/src/nvim_bridge.rs
@@ -241,9 +241,9 @@ pub enum CompletionItemKind {
     Variable,
 }
 
-impl From<&String> for CompletionItemKind {
-    fn from(from: &String) -> Self {
-        match from.as_str() {
+impl From<&str> for CompletionItemKind {
+    fn from(from: &str) -> Self {
+        match from {
             "class" => CompletionItemKind::Class,
             "color" => CompletionItemKind::Color,
             "constant" => CompletionItemKind::Constant,
@@ -752,9 +752,8 @@ fn parse_redraw_event(args: Vec<Value>) -> Vec<RedrawEvent> {
                     for item in unwrap_array!(args[0]) {
                         let item = unwrap_array!(item);
                         let word = unwrap_str!(item[0]).to_owned();
-                        let kind = CompletionItemKind::from(
-                            &unwrap_str!(item[1]).to_owned(),
-                        );
+                        let kind =
+                            CompletionItemKind::from(unwrap_str!(item[1]));
                         let kind_raw = unwrap_str!(item[1]).to_owned();
                         let menu = unwrap_str!(item[2]).to_owned();
                         let info = unwrap_str!(item[3]).to_owned();
--- a/src/nvim_bridge.rs
+++ b/src/nvim_bridge.rs
@@ -212,9 +212,78 @@ pub enum OptionSet {
 }
 
 #[derive(Clone)]
+pub enum CompletionItemKind {
+    Class,
+    Color,
+    Constant,
+    Constructor,
+    Enum,
+    EnumMember,
+    Event,
+    Function,
+    File,
+    Folder,
+    Field,
+    Interface,
+    Keyword,
+    Method,
+    Module,
+    Operator,
+    Property,
+    Reference,
+    Snippet,
+    Struct,
+    Text,
+    TypeParameter,
+    Unit,
+    Unknown,
+    Value,
+    Variable,
+}
+
+impl CompletionItemKind {
+    pub fn new_from_str(kind: &str) -> Self {
+        match kind {
+            "class" => CompletionItemKind::Class,
+            "color" => CompletionItemKind::Color,
+            "constant" => CompletionItemKind::Constant,
+            "constructor" => CompletionItemKind::Constructor,
+            "enum" => CompletionItemKind::Enum,
+            "enum member" => CompletionItemKind::EnumMember,
+            "event" => CompletionItemKind::Event,
+            "function" => CompletionItemKind::Function,
+            "file" => CompletionItemKind::File,
+            "folder" => CompletionItemKind::Folder,
+            "field" => CompletionItemKind::Field,
+            "interface" => CompletionItemKind::Interface,
+            "keyword" => CompletionItemKind::Keyword,
+            "method" => CompletionItemKind::Method,
+            "module" => CompletionItemKind::Module,
+            "operator" => CompletionItemKind::Operator,
+            "property" => CompletionItemKind::Property,
+            "reference" => CompletionItemKind::Reference,
+            "snippet" => CompletionItemKind::Snippet,
+            "struct" => CompletionItemKind::Struct,
+            "text" => CompletionItemKind::Text,
+            "type parameter" => CompletionItemKind::TypeParameter,
+            "unit" => CompletionItemKind::Unit,
+            "value" => CompletionItemKind::Value,
+            "variable" => CompletionItemKind::Variable,
+            _ => CompletionItemKind::Unknown,
+        }
+    }
+   pub fn is_unknown(&self) -> bool {
+        match self {
+            CompletionItemKind::Unknown => true,
+            _ => false,
+        }
+    }
+}
+
+#[derive(Clone)]
 pub struct CompletionItem {
     pub word: String,
-    pub kind: String,
+    pub kind: CompletionItemKind,
     pub menu: String,
     pub info: String,
 }
@@ -679,7 +748,7 @@ fn parse_redraw_event(args: Vec<Value>) -> Vec<RedrawEvent> {
                     for item in unwrap_array!(args[0]) {
                         let item = unwrap_array!(item);
                         let word = unwrap_str!(item[0]).to_owned();
-                        let kind = unwrap_str!(item[1]).to_owned();
+                        let kind = CompletionItemKind::new_from_str(&unwrap_str!(item[1]).to_owned());
                         let menu = unwrap_str!(item[2]).to_owned();
                         let info = unwrap_str!(item[3]).to_owned();
 
--- a/src/ui/popupmenu/completion_item_widget.rs
+++ b/src/ui/popupmenu/completion_item_widget.rs
@@ -1,7 +1,7 @@
 use gtk;
 use gtk::prelude::*;
 
-use nvim_bridge::CompletionItem;
+use nvim_bridge::{ CompletionItem, CompletionItemKind };
 use ui::color::Color;
 
 macro_rules! icon {
@@ -20,7 +20,9 @@ pub struct CompletionItemWidgetWrap {
     /// Label displaying `menu` for this item in the list.
     pub menu: gtk::Label,
     /// Image of the item in the row.
-    pub kind: gtk::Image,
+    pub image: gtk::Image,
+    /// Kind of the item
+    pub kind: CompletionItemKind,
     /// Root container.
     pub row: gtk::ListBoxRow,
 }
@@ -28,15 +30,31 @@ pub struct CompletionItemWidgetWrap {
 impl CompletionItemWidgetWrap {
     pub fn create(
         item: CompletionItem,
+        items: &Vec<CompletionItem>,
         css_provider: &gtk::CssProvider,
         icon_fg: &Color,
         size: f64,
-        know_items_kinds: bool,
     ) -> Self {
         let margin = (size / 3.0) as i32;
 
         let grid = gtk::Grid::new();
         grid.set_column_spacing(10);
+
+        let buf = get_icon_pixbuf(&item.kind, icon_fg, size);
+        let image = match &item.kind {
+            CompletionItemKind::Unknown => {
+                // If another item in the popupmenu has a known kind,
+                // give kind_image an icon.
+                if items.iter().any(|item| !item.kind.is_unknown()) {
+                    gtk::Image::new_from_pixbuf(&buf) 
+                } else { gtk::Image::new() }
+            }
+            _ => gtk::Image::new_from_pixbuf(&buf),
+        };
+
+        // kind_image.set_tooltip_text(format!("kind: '{}'", &item.kind.get_kind()).as_str());
+        image.set_margin_start(margin);
+        grid.attach(&image, 0, 0, 1, 1);
 
         let menu = gtk::Label::new(item.menu.as_str());
         menu.set_halign(gtk::Align::End);
@@ -67,36 +85,17 @@ impl CompletionItemWidgetWrap {
         let row = gtk::ListBoxRow::new();
         row.add(&grid);
 
-        let buf = get_icon_pixbuf(&item.kind.as_str(),
-                                    icon_fg, 
-                                    size,
-                                    know_items_kinds);
-        match buf {
-            Ok(buff) => {
-                let kind = gtk::Image::new_from_pixbuf(&buff);
-                kind.set_tooltip_text(format!("kind: '{}'", item.kind).as_str());
-                kind.set_margin_start(margin);
-                grid.attach(&kind, 0, 0, 1, 1);
-                add_css_provider!(css_provider, grid, word, kind, info, row, menu);
-                return CompletionItemWidgetWrap {
-                    item,
-                    info,
-                    row,
-                    kind,
-                    menu,
-                };
-            },
-            Err(_) => {
-                add_css_provider!(css_provider, grid, word, info, row, menu);
-                return CompletionItemWidgetWrap {
-                    item,
-                    info,
-                    row,
-                    kind: gtk::Image::new(),
-                    menu,
-                }
-            },
-        };
+        add_css_provider!(css_provider, grid, word, image, info, row, menu);
+
+        let kind = item.kind.clone();
+        CompletionItemWidgetWrap {
+            item,
+            info,
+            row,
+            image,
+            kind,
+            menu,
+        }
     }
 }
 
@@ -108,69 +107,64 @@ fn shorten_info(info: &String) -> String {
 }
 
 pub fn get_icon_pixbuf(
-    kind: &str,
+    kind: &CompletionItemKind,
     color: &Color,
     size: f64,
-    know_items_kinds: bool,
-) -> Result<gdk_pixbuf::Pixbuf, gdk_pixbuf::Error> {
-    let contents = get_icon_name_for_kind(kind,
-                                          &color,
-                                          size,
-                                          know_items_kinds);
+) -> gdk_pixbuf::Pixbuf {
+    let contents = get_icon_name_for_kind(kind, &color, size);
     let stream = gio::MemoryInputStream::new_from_bytes(&glib::Bytes::from(
         contents.as_bytes(),
     ));
-    let buf = gdk_pixbuf::Pixbuf::new_from_stream(&stream, None);
+    let buf = gdk_pixbuf::Pixbuf::new_from_stream(&stream, None).unwrap();
 
     buf
 }
 
-fn get_icon_name_for_kind(kind: &str, color: &Color, size: f64,
-                          know_items_kinds: bool) -> String {
+fn get_icon_name_for_kind(
+    kind: &CompletionItemKind,
+    color: &Color,
+    size: f64,
+) -> String {
     let color = color.to_hex();
 
     let size = size * 1.1;
 
+    use self::CompletionItemKind::*;
     match kind {
-        "method" | "function" | "constructor" => {
-            icon!("../../../assets/icons/box.svg", color, size)
-        }
-        "field" => {
+        Constructor => icon!("../../../assets/icons/box.svg", color, size),
+        Method => icon!("../../../assets/icons/box.svg", color, size),
+        Function => icon!("../../../assets/icons/box.svg", color, size),
+        Field => {
             icon!("../../../assets/icons/chevrons-right.svg", color, size)
         }
-        "event" => icon!("../../../assets/icons/zap.svg", color, size),
-        "operator" => icon!("../../../assets/icons/sliders.svg", color, size),
-        "variable" => icon!("../../../assets/icons/disc.svg", color, size),
-        "class" => icon!("../../../assets/icons/share-2.svg", color, size),
-        "interface" => {
+        Event => icon!("../../../assets/icons/zap.svg", color, size),
+        Operator => icon!("../../../assets/icons/sliders.svg", color, size),
+        Variable => icon!("../../../assets/icons/disc.svg", color, size),
+        Class => icon!("../../../assets/icons/share-2.svg", color, size),
+        Interface => {
             icon!("../../../assets/icons/book-open.svg", color, size)
         }
-        "struct" => icon!("../../../assets/icons/align-left.svg", color, size),
-        "type parameter" => {
+        Struct => icon!("../../../assets/icons/align-left.svg", color, size),
+        TypeParameter => {
             icon!("../../../assets/icons/type.svg", color, size)
         }
-        "module" => icon!("../../../assets/icons/code.svg", color, size),
-        "property" => icon!("../../../assets/icons/key.svg", color, size),
-        "unit" => icon!("../../../assets/icons/compass.svg", color, size),
-        "constant" => icon!("../../../assets/icons/shield.svg", color, size),
-        "value" | "enum" => {
+        Module => icon!("../../../assets/icons/code.svg", color, size),
+        Property => icon!("../../../assets/icons/key.svg", color, size),
+        Unit => icon!("../../../assets/icons/compass.svg", color, size),
+        Constant => icon!("../../../assets/icons/shield.svg", color, size),
+        Value => icon!("../../../assets/icons/database.svg", color, size),
+        Enum => {
             icon!("../../../assets/icons/database.svg", color, size)
         }
-        "enum member" => icon!("../../../assets/icons/tag.svg", color, size),
-        "keyword" => icon!("../../../assets/icons/link-2.svg", color, size),
-        "text" => icon!("../../../assets/icons/at-sign.svg", color, size),
-        "color" => icon!("../../../assets/icons/aperture.svg", color, size),
-        "file" => icon!("../../../assets/icons/file.svg", color, size),
-        "reference" => icon!("../../../assets/icons/link.svg", color, size),
-        "snippet" => icon!("../../../assets/icons/file-text.svg", color, size),
-        "folder" => icon!("../../../assets/icons/folder.svg", color, size),
+        EnumMember => icon!("../../../assets/icons/tag.svg", color, size),
+        Keyword => icon!("../../../assets/icons/link-2.svg", color, size),
+        Text => icon!("../../../assets/icons/at-sign.svg", color, size),
+        Color => icon!("../../../assets/icons/aperture.svg", color, size),
+        File => icon!("../../../assets/icons/file.svg", color, size),
+        Reference => icon!("../../../assets/icons/link.svg", color, size),
+        Snippet => icon!("../../../assets/icons/file-text.svg", color, size),
+        Folder => icon!("../../../assets/icons/folder.svg", color, size),
 
-        _ => {
-            if know_items_kinds {
-                icon!("../../../assets/icons/help-circle.svg", color, size)
-            } else {
-                String::from("")
-            }
-        },
+        _ => icon!("../../../assets/icons/help-circle.svg", color, size),
     }
 }
--- a/src/ui/popupmenu/completion_item_widget.rs
+++ b/src/ui/popupmenu/completion_item_widget.rs
@@ -156,7 +156,6 @@ fn get_icon_name_for_kind(kind: &str, color: &Color, size: f64) -> String {
         "snippet" => icon!("../../../assets/icons/file-text.svg", color, size),
         "folder" => icon!("../../../assets/icons/folder.svg", color, size),
 
-       // _ => icon!("../../../assets/icons/help-circle.svg", color, size),
         _ => String::from("None"),
     }
 }
--- a/src/ui/popupmenu/completion_item_widget.rs
+++ b/src/ui/popupmenu/completion_item_widget.rs
@@ -40,19 +40,11 @@ impl CompletionItemWidgetWrap {
         let grid = gtk::Grid::new();
         grid.set_column_spacing(10);
 
-        let buf = get_icon_pixbuf(&item.kind, icon_fg, size);
-        let image = match &item.kind {
-            CompletionItemKind::Unknown => {
-                // If another item in the popupmenu has a known kind,
-                // give kind_image an icon.
-                if show_kind {
-                    gtk::Image::new_from_pixbuf(&buf)
-                } else {
-                    gtk::Image::new()
-                }
-            }
-            _ => gtk::Image::new_from_pixbuf(&buf),
-        };
+        let image = gtk::Image::new();
+        if show_kind {
+            let buf = get_icon_pixbuf(&item.kind, icon_fg, size);
+            image.set_from_pixbuf(&buf);
+        }
 
         image.set_tooltip_text(format!("kind: '{}'", item.kind_raw).as_str());
         image.set_margin_start(margin);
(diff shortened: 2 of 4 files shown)


badosu:
I think we should hide icons only when all are unknown, otherwise it will introduce visual inconsistency when LSP kinds are supported but customized or not provided for a few items.

badosu:
Unrelated to the issue, but relevant to your use case

I haven't had the time to figure out the kind names [here](https://github.com/autozimu/LanguageClient-neovim/blob/next/rplugin/python3/denite/lsp/protocol.py), but if you add `dbg!(kind)` [here](https://github.com/badosu/gnvim/blob/ba22b215cd378fbf764a3b8ecbf58a88584eac98/src/ui/popupmenu/completion_item_widget.rs#L112) implementing support for it is pretty trivial: https://github.com/vhakulinen/gnvim/pull/49

smolck:
Good point! So we find a different icon for text completions, and then if all of the completions are unknown, we show no icons. I actually got this functionality to work in a local clone of mine. It uses a boolean parameter in the `create`, `get_icon_name_for_kind` and `get_icon_pixbuf` functions in completion_item_widget.rs. Should I push it to the PR for review?

smolck:
Now the pmenu looks like this if any of the completions are known (note the bottom unknown one):
image

And this if none are known:
image

However, I used the help-circle icon for a placeholder. Any ideas for a different icon for text/unknown completions? (Or should we keep the help-circle?)

src/ui/popupmenu/completion_item_widget.rs
```diff
@@ -144,6 +165,12 @@ fn get_icon_name_for_kind(kind: &str, color: &Color, size: f64) -> String {
         "snippet" => icon!("../../../assets/icons/file-text.svg", color, size),
         "folder" => icon!("../../../assets/icons/folder.svg", color, size),
 
-        _ => icon!("../../../assets/icons/help-circle.svg", color, size),
+        _ => {
+            if know_items_kinds {
+                icon!("../../../assets/icons/help-circle.svg", color, size)
+            } else {
+                String::from("")
```
badosu: What do you think of always returning `String::from("")` or something that's easy to identify as not matched (e.g. `None`).

And then on the place that assigns icons, doing something like this (I am not too familiar with rust, pseudocode):

```
// Get all are not matched ahead of time
let any_kind_matches = all(state.items_to_load, (i) => Some(get_icon_pixbuf(&i.kind, .., ..))

// ...
// specific item
if any_kind_matches {
  prev.kind.set_from_pixbuf(&buf.unwrap());
}
```

This way we void passing too much state for this, just getting the variable once. I hope I was able to communicate clearly, let me know if that's not the case.

smolck: So you're saying that we should do the checks for which icon should be used outside of the `get_icon_name_for_kind` and `get_icon_pixbuf` functions?

If that's what you mean (correct me if I'm wrong), then I totally agree. We could instead create a function that takes all of the popupmenu's CompletionItems as a parameter (in addition to all of the parameters of `get_icon_pixbuf`), do the needed checks inside of that function to determine whether or not a help-circle (or another, TBD icon) should be used, and then return the necessary pixbuf. Thus, we don't needlessly pass around too much state.

Is that what you were thinking, or did I misunderstand?

badosu: > @badosu So you're saying that we should do the checks for which icon should be used outside of the get_icon_name_for_kind and get_icon_pixbuf functions?

No, what I am trying to convey is to separate the case where no match was found from the `get_icon_name_for_kind`. So if it's not found we can just make a check to see if it was not found for all items. If that's the case then we can decide if we set the question mark icon or not (whether all items are matched or not).

The idea here is that we shouldn't need to pass the information if all items are matched (`know_items_kinds`) inside any other scope than the one where we have all items and the one that sets the pixbuf.


I tried coding this, but was having some issues with the `icon!` macro that prevented me from proceeding forward.

smolck: Ok, that makes sense, and I like it. In order to implement it, however, I feel as though `get_icon_pixbuf` gets in the way. What do you think about making `get_icon_name_for_kind` public and then calling it directly? If `get_icon_name_for_kind` returned "None" or an empty string we would then check the strings returned for all of the other items to see if they are "None" or empty likewise. Afterwards we call `get_icon_pixbuf` with the necessary `contents` parameter (help-circle icon if we need to use an icon). We could probably not call `get_icon_pixbuf` at all in `CompletionItemWidgetWrap::create()` and use an empty image instead if we didn't need the help-circle.

badosu: I am still a beginner on Rust, so take my suggestions with a grain of salt :-).

smolck: It’s all good (I’m still learning the lang. too 😉), and your suggestions have been great. So you want me to push the change I stated above? (Also, check out Rust’s “book” if you haven’t already and get the chance, great doc for Rust.)

smolck: In order to check the other items' kinds in the `CompletionItemWidgetWrap::create()` function, I would need to pass it the other CompletionItems as a parameter (instead of `know_items_kinds`). Sound good?

vhakulinen:
Showing the help-circle when there are _some_ known kinds is the way to go, so we avoid visual inconsistencies (just like @badosu noted). On a case when there are no icons shown (e.g. all kinds are unknown), remember to add some padding to the left side of the actual text. I would also avoid differing the code paths too much.

Regarding the issue that you are still facing: how to detect when a `kind` is known or not. I would try this: In `nvim_bridge`, make the completion item's kind be type of `CompletionItemKind` which would be an enum (that would look something like this):

```rust
pub enum CompletionItemKind {
    Function,
    Constructor,
    File,
    Folder,
    //...
    Unknown,
}

impl CompletionItemKind {
    pub fn is_unknown(&self) -> bool {
        match self {
            CompletionItemKind::Unknown => true,
            _ => false,
        }
    }
}

// Detecting if some of `items` is unknown/known:
items.iter().find(|item| !item.is_unkown()).is_some())
```
I didn't try this, so I'm not completely sure how it'll play out. We already have a "know set" of kinds, might has well make the type system aware of it.

I'll try to add some notes to other parts of the code some other day (its getting late here).

smolck:
Thanks @vhakulinen for the info, I will try to get that working. If it goes well I’ll push a commit for review.

smolck:
So, how does this look for nvim_bridge.rs? Is this what you were thinking @vhakulinen?

```diff
diff --git a/src/nvim_bridge.rs b/src/nvim_bridge.rs
index 72a9777..fe0eec5 100644
--- a/src/nvim_bridge.rs
+++ b/src/nvim_bridge.rs
@@ -211,10 +211,79 @@ pub enum OptionSet {
     NotSupported(String),
 }
 
+#[derive(Clone)]
+pub enum CompletionItemKind {
+    Class,
+    Color,
+    Constant,
+    Constructor,
+    Enum,
+    EnumMember,
+    Event,
+    Function,
+    File,
+    Folder,
+    Field,
+    Interface,
+    Keyword,
+    Method,
+    Module,
+    Operator,
+    Property,
+    Reference,
+    Snippet,
+    Struct,
+    Text,
+    TypeParameter,
+    Unit,
+    Unknown,
+    Value,
+    Variable,
+}
+
+impl CompletionItemKind {
+    pub fn new_from_str(kind: &str) -> Self {
+        match kind {
+            "class" => CompletionItemKind::Class,
+            "color" => CompletionItemKind::Color,
+            "constant" => CompletionItemKind::Constant,
+            "constructor" => CompletionItemKind::Constructor,
+            "enum" => CompletionItemKind::Enum,
+            "enum member" => CompletionItemKind::EnumMember,
+            "event" => CompletionItemKind::Event,
+            "function" => CompletionItemKind::Function,
+            "file" => CompletionItemKind::File,
+            "folder" => CompletionItemKind::Folder,
+            "field" => CompletionItemKind::Field,
+            "interface" => CompletionItemKind::Interface,
+            "keyword" => CompletionItemKind::Keyword,
+            "method" => CompletionItemKind::Method,
+            "module" => CompletionItemKind::Module,
+            "operator" => CompletionItemKind::Operator,
+            "property" => CompletionItemKind::Property,
+            "reference" => CompletionItemKind::Reference,
+            "snippet" => CompletionItemKind::Snippet,
+            "struct" => CompletionItemKind::Struct,
+            "text" => CompletionItemKind::Text,
+            "type parameter" => CompletionItemKind::TypeParameter,
+            "unit" => CompletionItemKind::Unit,
+            "value" => CompletionItemKind::Value,
+            "variable" => CompletionItemKind::Variable,
+            _ => CompletionItemKind::Unknown,
+        }
+    }
+   pub fn is_unknown(&self) -> bool {
+        match self {
+            CompletionItemKind::Unknown => true,
+            _ => false,
+        }
+    }
+}
+
 #[derive(Clone)]
 pub struct CompletionItem {
     pub word: String,
-    pub kind: String,
+    pub kind: CompletionItemKind,
     pub menu: String,
     pub info: String,
 }
@@ -679,7 +748,7 @@ fn parse_redraw_event(args: Vec<Value>) -> Vec<RedrawEvent> {
                     for item in unwrap_array!(args[0]) {
                         let item = unwrap_array!(item);
                         let word = unwrap_str!(item[0]).to_owned();
-                        let kind = unwrap_str!(item[1]).to_owned();
+                        let kind = CompletionItemKind::new_from_str(&unwrap_str!(item[1]).to_owned());
                         let menu = unwrap_str!(item[2]).to_owned();
                         let info = unwrap_str!(item[3]).to_owned();
```

I made changes in other files to accommodate this diff, but this is the core of the change, so I want to make sure this is solid before I settle on my other diffs (and put them up for review).

smolck:
One question: should I implement a `get_kind` function (in the enum) that returns the corresponding string of a `CompletionItemKind`? The reason I ask is because of the `set_tooltip_text` function called in the `CompletionItemWidgetWrap::create` in completion_item_widget.rs. The thing is, I don’t want to put needless logic in that function, but I also don’t want to do the same for the enum if it will only be used once. Should I make a function that does that in completion_item_widget.rs, or do something else entirely?

smolck:
Here is part of the current (unfinished) diff for completion_item_widget.rs. Note the additional and changed variables in `CompletionItemWidgetWrap` (done for use in popupmenu.rs), and the padding added to each `CompletionItemWidgetWrap`'s image regardless of whether or not an icon is shown:
```diff
diff --git a/src/ui/popupmenu/completion_item_widget.rs b/src/ui/popupmenu/completion_item_widget.rs
index 5b9f8f7..4626771 100644
--- a/src/ui/popupmenu/completion_item_widget.rs
+++ b/src/ui/popupmenu/completion_item_widget.rs
@@ -1,7 +1,7 @@
 use gtk;
 use gtk::prelude::*;
 
-use nvim_bridge::CompletionItem;
+use nvim_bridge::{ CompletionItem, CompletionItemKind };
 use ui::color::Color;
 
 macro_rules! icon {
@@ -20,7 +20,9 @@ pub struct CompletionItemWidgetWrap {
     /// Label displaying `menu` for this item in the list.
     pub menu: gtk::Label,
     /// Image of the item in the row.
-    pub kind: gtk::Image,
+    pub image: gtk::Image,
+    /// Kind of the item
+    pub kind: CompletionItemKind,
     /// Root container.
     pub row: gtk::ListBoxRow,
 }
@@ -28,16 +30,32 @@ pub struct CompletionItemWidgetWrap {
 impl CompletionItemWidgetWrap {
     pub fn create(
         item: CompletionItem,
+        items: &Vec<CompletionItem>,
         css_provider: &gtk::CssProvider,
         icon_fg: &Color,
         size: f64,
-        know_items_kinds: bool,
     ) -> Self {
         let margin = (size / 3.0) as i32;
 
         let grid = gtk::Grid::new();
         grid.set_column_spacing(10);
 
+        let buf = get_icon_pixbuf(&item.kind, icon_fg, size);
+        let image = match &item.kind {
+            CompletionItemKind::Unknown => {
+                // If another item in the popupmenu has a known kind,
+                // give kind_image an icon.
+                if items.iter().any(|item| !item.kind.is_unknown()) {
+                    gtk::Image::new_from_pixbuf(&buf) 
+                } else { gtk::Image::new() }
+            }
+            _ => gtk::Image::new_from_pixbuf(&buf),
+        };
+
+        // kind_image.set_tooltip_text(format!("kind: '{}'", &item.kind.get_kind()).as_str());
+        image.set_margin_start(margin);
+        grid.attach(&image, 0, 0, 1, 1);
+
         let menu = gtk::Label::new(item.menu.as_str());
         menu.set_halign(gtk::Align::End);
         menu.set_hexpand(true);
@@ -67,36 +85,17 @@ impl CompletionItemWidgetWrap {
         let row = gtk::ListBoxRow::new();
         row.add(&grid);
 
-        let buf = get_icon_pixbuf(&item.kind.as_str(),
-                                    icon_fg, 
-                                    size,
-                                    know_items_kinds);
-        match buf {
-            Ok(buff) => {
-                let kind = gtk::Image::new_from_pixbuf(&buff);
-                kind.set_tooltip_text(format!("kind: '{}'", item.kind).as_str());
-                kind.set_margin_start(margin);
-                grid.attach(&kind, 0, 0, 1, 1);
-                add_css_provider!(css_provider, grid, word, kind, info, row, menu);
-                return CompletionItemWidgetWrap {
-                    item,
-                    info,
-                    row,
-                    kind,
-                    menu,
-                };
-            },
-            Err(_) => {
-                add_css_provider!(css_provider, grid, word, info, row, menu);
-                return CompletionItemWidgetWrap {
-                    item,
-                    info,
-                    row,
-                    kind: gtk::Image::new(),
-                    menu,
-                }
-            },
-        };
+        add_css_provider!(css_provider, grid, word, image, info, row, menu);
+
+        let kind = item.kind.clone();
+        CompletionItemWidgetWrap {
+            item,
+            info,
+            row,
+            image,
+            kind,
+            menu,
+        }
     }
}
```
I also made `get_icon_pixbuf` and `get_icon_name_for_kind` take `&CompletionItemKind` instead of `&str` for their `kind` parameter (I matched the `&CompletionItemKind` in `get_icon_name_for_kind` instead of `&str`). I can push these changes and the others, or post them in a comment, whichever is preferred. (For now and the future, which would you prefer me do? Post the changes in comments before pushing them, or just pushing them?)

smolck:
Changes have been pushed for review. Suggestions are welcome as always, and also note that I have commented out the `set_tooltip_text` function in `CompletionItemWidgetWrap::create` for now, until we know what to do there.

src/ui/popupmenu/completion_item_widget.rs
```diff
@@ -37,11 +40,21 @@ impl CompletionItemWidgetWrap {
         let grid = gtk::Grid::new();
         grid.set_column_spacing(10);
 
-        let buf = get_icon_pixbuf(&item.kind.as_str(), icon_fg, size);
-        let kind = gtk::Image::new_from_pixbuf(&buf);
-        kind.set_tooltip_text(format!("kind: '{}'", item.kind).as_str());
-        kind.set_margin_start(margin);
-        grid.attach(&kind, 0, 0, 1, 1);
+        let buf = get_icon_pixbuf(&item.kind, icon_fg, size);
```
badosu: I see you refactored to avoid passing `all_kinds_known` to other scopes, good one! :clap:

src/nvim_bridge.rs
```diff
@@ -211,10 +211,79 @@ pub enum OptionSet {
...
+    Unit,
+    Unknown,
+    Value,
+    Variable,
+}
+
+impl CompletionItemKind {
+    pub fn new_from_str(kind: &str) -> Self {
+        match kind {
+            "class" => CompletionItemKind::Class,
```
badosu: This will make other LSP's support much easier.

src/ui/popupmenu/popupmenu.rs
```diff
@@ -442,10 +444,12 @@ impl Popupmenu {
                     *id.borrow_mut() = Some(sig_id);
                 }
 
-                // Update the `kind` icon with "selected" fg color.
-                let buf =
-                    get_icon_pixbuf(&item.item.kind, &fg_sel, font_height);
-                item.kind.set_from_pixbuf(&buf);
+                // Update the `kind` icon with "selected" fg color if it is an icon.
+                let buf = get_icon_pixbuf(&item.item.kind, &fg_sel, font_height);
+                if items.iter().any(|item| !item.kind.is_unknown()) {
```
badosu: I wonder if we can assign this to a variable to avoid running this two times, here and on line 393?

We would have to move line 398 and the check below it before the assignment.

Just a nitpick

src/ui/popupmenu/completion_item_widget.rs
```diff
@@ -37,11 +40,21 @@ impl CompletionItemWidgetWrap {
...
+        let image = match &item.kind {
+            CompletionItemKind::Unknown => {
+                // If another item in the popupmenu has a known kind,
+                // give kind_image an icon.
+                if items.iter().any(|item| !item.kind.is_unknown()) {
```
badosu: This is being ran multiple times while it can be assigned to a variable just once, perhaps before line 44

src/ui/popupmenu/popupmenu.rs
smolck: Do you mean like this?
```diff
diff --git a/src/ui/popupmenu/popupmenu.rs b/src/ui/popupmenu/popupmenu.rs
index b4f5248..9425f3c 100644
--- a/src/ui/popupmenu/popupmenu.rs
+++ b/src/ui/popupmenu/popupmenu.rs
@@ -383,6 +383,7 @@ impl Popupmenu {
 
         self.items.once_loaded(Some(item_num), move |items| {
             let mut state = state.borrow_mut();
+            let items_have_kinds = items.iter().any(|item| !item.kind.is_unknown());
 
             if let Some(prev) = items.get(state.selected as usize) {
                 prev.info.set_visible(false);
@@ -390,9 +391,7 @@ impl Popupmenu {
 
                 // Update the `kind` icon with default fg color if it is an icon.
                 let buf = get_icon_pixbuf(&prev.item.kind, &fg, font_height);
-                if items.iter().any(|item| !item.kind.is_unknown())  {
-                    prev.image.set_from_pixbuf(&buf);
-                }
+                if items_have_kinds { prev.image.set_from_pixbuf(&buf) }
             }
 
             state.selected = item_num;
@@ -446,10 +445,8 @@ impl Popupmenu {
 
                 // Update the `kind` icon with "selected" fg color if it is an icon.
                 let buf = get_icon_pixbuf(&item.item.kind, &fg_sel, font_height);
-                if items.iter().any(|item| !item.kind.is_unknown()) {
-                    item.image.set_from_pixbuf(&buf);
-                }
-                
+                if items_have_kinds { item.image.set_from_pixbuf(&buf) }
+               
                 let newline =
                     if item.item.menu.len() > 0 && item.item.info.len() > 0 {
                         "\n"
```
The only problem here is it happens even if `items_have_kinds` isn't used. (Also, I am not sure on the variable name)

badosu: The only place where it's not used is on https://github.com/vhakulinen/gnvim/blob/5bbccf3d0328cf31ed60a616f51de1a29eba492c/src/ui/popupmenu/popupmenu.rs#L398-L413

Yeah, it seems it's not a good idea to touch this, it's the code path when the popup menu is just opened, nevermind.

smolck: So leave it as is?

src/nvim_bridge.rs
smolck: @badosu if we wanted to we could go ahead and add the support for the LSPs here, since this is where it matches the kind.

src/ui/popupmenu/completion_item_widget.rs
```diff
@@ -20,14 +20,17 @@ pub struct CompletionItemWidgetWrap {
...
+    /// Kind of the item
+    pub kind: CompletionItemKind,
     /// Root container.
     pub row: gtk::ListBoxRow,
 }
 
 impl CompletionItemWidgetWrap {
     pub fn create(
         item: CompletionItem,
+        items: &Vec<CompletionItem>,
```
vhakulinen: Here, you should take parameter `show_kind` which is a boolean and tells if the kind image should be shown or not. Currently, either all the kind images are shown, or they are not.

src/ui/popupmenu/completion_item_widget.rs
```diff
@@ -37,11 +40,21 @@ impl CompletionItemWidgetWrap {
         let grid = gtk::Grid::new();
         grid.set_column_spacing(10);
 
-        let buf = get_icon_pixbuf(&item.kind.as_str(), icon_fg, size);
-        let kind = gtk::Image::new_from_pixbuf(&buf);
-        kind.set_tooltip_text(format!("kind: '{}'", item.kind).as_str());
-        kind.set_margin_start(margin);
-        grid.attach(&kind, 0, 0, 1, 1);
+        let buf = get_icon_pixbuf(&item.kind, icon_fg, size);
+        let image = match &item.kind {
```
vhakulinen: Here, you can just use `show_kind` which would be passed in as a parameter.

src/ui/popupmenu/completion_item_widget.rs
```diff
@@ -72,12 +85,14 @@ impl CompletionItemWidgetWrap {
         let row = gtk::ListBoxRow::new();
         row.add(&grid);
 
-        add_css_provider!(css_provider, grid, kind, word, info, row, menu);
+        add_css_provider!(css_provider, grid, word, image, info, row, menu);
 
+        let kind = item.kind.clone();
```
vhakulinen: Instead of doing a implicit clone, require a owner ship of the value and do the clone on a higher level.

src/ui/popupmenu/lazy_loader.rs
```diff
@@ -67,7 +67,7 @@ impl LazyLoader {
         let mut state = self.state.borrow_mut();
         state.clear();
 
-        state.items_to_load = items;
+        state.items_to_load = items.clone();
```
vhakulinen: Calculate the `show_kind` value here, and if needed, store that to the `State` struct.

src/ui/popupmenu/popupmenu.rs
```diff
@@ -388,9 +388,11 @@ impl Popupmenu {
                 prev.info.set_visible(false);
                 prev.menu.set_visible(false);
 
-                // Update the `kind` icon with default fg color.
+                // Update the `kind` icon with default fg color if it is an icon.
                 let buf = get_icon_pixbuf(&prev.item.kind, &fg, font_height);
-                prev.kind.set_from_pixbuf(&buf);
+                if items.iter().any(|item| !item.kind.is_unknown())  {
```
vhakulinen: Here, you can use `state.show_kind` once that is added.

src/ui/popupmenu/popupmenu.rs
vhakulinen: Here too you can use `state.show_kind` once that is added.

vhakulinen:
Remember to run `cargo fmt`.

src/ui/popupmenu/completion_item_widget.rs
smolck: @vhakulinen Do you mean using `to_owned()` instead, or doing something else? (I am not entirely familiar with the way cloning works, so that's why I'm asking.)

src/ui/popupmenu/popupmenu.rs
smolck: @vhakulinen Do you want me to add `show_kind` to the _popupmenu's_ `State` struct, or is there another way to use `state.show_kind` here that doesn't require adding that?

src/ui/popupmenu/completion_item_widget.rs
vhakulinen: Oh, sorry. This one was a brainfart on my part. Ignore this comment.

src/ui/popupmenu/completion_item_widget.rs
```diff
@@ -37,11 +40,21 @@ impl CompletionItemWidgetWrap {
...
+                // If another item in the popupmenu has a known kind,
+                // give kind_image an icon.
+                if items.iter().any(|item| !item.kind.is_unknown()) {
+                    gtk::Image::new_from_pixbuf(&buf) 
+                } else { gtk::Image::new() }
+            }
+            _ => gtk::Image::new_from_pixbuf(&buf),
+        };
+
+        // kind_image.set_tooltip_text(format!("kind: '{}'", &item.kind.get_kind()).as_str());
```
smolck: @vhakulinen What should I do here about setting the tooltip text for the image?

src/ui/popupmenu/popupmenu.rs
vhakulinen: Ah, right. The `show_kind` should go into the `LazyLoader`'s state. You'll have to add an function to `LazyLoader` to then get it.

src/ui/popupmenu/completion_item_widget.rs
vhakulinen: I think you can just add `kind_raw` to the `CompletionItem` struct and then use that to add the tooltip text. The tooltip text is here for debugging/troubleshooting.

src/nvim_bridge.rs
```diff
@@ -211,10 +211,79 @@ pub enum OptionSet {
...
+    Text,
+    TypeParameter,
+    Unit,
+    Unknown,
+    Value,
+    Variable,
+}
+
+impl CompletionItemKind {
+    pub fn new_from_str(kind: &str) -> Self {
```
smolck: @vhakulinen Should I impl the `From<&str>` trait here instead of creating the `new_from_str` function?

vhakulinen: No need. You can do that ofc if you want to.

smolck:
Changes have been pushed for review.

src/nvim_bridge.rs
```diff
@@ -211,10 +211,83 @@ pub enum OptionSet {
...
+    Struct,
+    Text,
+    TypeParameter,
+    Unit,
+    Unknown,
+    Value,
+    Variable,
+}
+
+impl From<&String> for CompletionItemKind {
```
vhakulinen: This should be `From<&str>` instead of `From<&String>`.

src/ui/popupmenu/completion_item_widget.rs
```diff
@@ -37,11 +40,23 @@ impl CompletionItemWidgetWrap {
         let grid = gtk::Grid::new();
         grid.set_column_spacing(10);
 
-        let buf = get_icon_pixbuf(&item.kind.as_str(), icon_fg, size);
-        let kind = gtk::Image::new_from_pixbuf(&buf);
-        kind.set_tooltip_text(format!("kind: '{}'", item.kind).as_str());
-        kind.set_margin_start(margin);
-        grid.attach(&kind, 0, 0, 1, 1);
+        let buf = get_icon_pixbuf(&item.kind, icon_fg, size);
+        let image = match &item.kind {
```
vhakulinen: Wouldn't this do the trick?
```rust
let image = gtk::Image::new();
if show_kind {
    let buf = ....;
    image.set_from_pixbuf(&buf);
}
```

We either show the kind or we don't - no matter what the kind type is.

src/ui/popupmenu/popupmenu.rs
```diff
@@ -388,9 +389,11 @@ impl Popupmenu {
                 prev.info.set_visible(false);
                 prev.menu.set_visible(false);
 
-                // Update the `kind` icon with default fg color.
+                // Update the `kind` icon with default fg color if it is an icon.
                 let buf = get_icon_pixbuf(&prev.item.kind, &fg, font_height);
```
vhakulinen: Move this inside the if clause. Also revert the documentation to the original form.

src/ui/popupmenu/popupmenu.rs
```diff
@@ -442,10 +445,13 @@ impl Popupmenu {
                     *id.borrow_mut() = Some(sig_id);
                 }
 
-                // Update the `kind` icon with "selected" fg color.
+                // Update the `kind` icon with "selected" fg color if it is an icon.
                 let buf =
                     get_icon_pixbuf(&item.item.kind, &fg_sel, font_height);
-                item.kind.set_from_pixbuf(&buf);
+                if show_kind {
```
vhakulinen: Same here, move the buf creation inside the if block and revert the comment.

vhakulinen:
Looking good! Just couple minor things that need to be fixed.

src/ui/popupmenu/completion_item_widget.rs
smolck: Great point, didn't think about that. Thanks!

src/nvim_bridge.rs
smolck: @vhakulinen After that change, I would need to do something like this in nvim_bridge.rs:
```diff
@@ -753,7 +753,7 @@ fn parse_redraw_event(args: Vec<Value>) -> Vec<RedrawEvent> {
                         let item = unwrap_array!(item);
                         let word = unwrap_str!(item[0]).to_owned();
                         let kind = CompletionItemKind::from(
-                            &unwrap_str!(item[1]).to_owned(),
+                            unwrap_str!(item[1]),
                         );
                         let kind_raw = unwrap_str!(item[1]).to_owned();
                         let menu = unwrap_str!(item[2]).to_owned();
```
Will the above work as intended?

smolck:
@vhakulinen I’ve pushed the changes you requested. If there is anything left that needs fixing/changing please let me know.

As a side note, thank you for all of your help with this PR and my previous one. Taught (and still teaching) me valuable information about Rust that I didn’t know, as well as better, more efficient and idiomatic ways of doing things.

smolck:
This is a little strange to me. In the below case, the behavior is as expected:
image
in that the bottom, unknown completion is in line with the others. However, here that is not the case:
image
Notice how the bottom (unknown) completion is out of line with the others. I don't know if this is important enough to need fixing, but wanted to point it out nonetheless. If anyone knows how to fix this or what's causing it, please let me know.

Also, here is how it looks when all completions are unknown, just as an FYI:
image

badosu:
Try adding a `dbg!(item.menu)` and seeing if the text itself coming from the LSP is padded.

src/ui/popupmenu/completion_item_widget.rs
```diff
@@ -37,11 +40,15 @@ impl CompletionItemWidgetWrap {
...
-        kind.set_margin_start(margin);
-        grid.attach(&kind, 0, 0, 1, 1);
+        let image = gtk::Image::new();
+        if show_kind {
+            let buf = get_icon_pixbuf(&item.kind, icon_fg, size);
+            image.set_from_pixbuf(&buf);
+        }
+
+        image.set_tooltip_text(format!("kind: '{}'", item.kind_raw).as_str());
+        image.set_margin_start(margin);
```
badosu: I think moving this to the `show_kind` branch above should fix the padding issue.

smolck: @badosu It appears that the padding problem is coming directly from the LSP. Here is the output from `dbg!(&item.word)` (notice the lack of a space in the bottom completion string):

image

So I guess we can just not worry about it then, right? (It's from the LSP? Correct me if I'm wrong.)

> Also the surrounding lines, it should not be necessary to add it to the grid or set tooltip info.

I didn't quite understand this. Do you mean if `show_kind` is false, we don't need to call those functions (`set_tooltip_text` and `grid.attach`)?

badosu: This is for the case where all unknown items show margin. There's no reason to attach images to the popup menu when that's the case, unless if it breaks some other structure that depends on their existence.

badosu: > @badosu It appears that the padding problem may be coming directly from the LSP (vim-lsp). Here is the output from dbg!(&item.word) (notice the lack of a space in the bottom completion string):

Also, it may not necessarily be the fault of the LSP, you need to inspect as far up on the stack as possible to confirm if we aren't changing the text by mistake (since this was probably not happening before).

Try debugging `popupmenu.items` [here](https://github.com/badosu/gnvim/blob/4b248cda4a6858544c6dbe9a4ad79796a7008247/src/ui/ui.rs#L676).

smolck: @badosu I used `dbg!(&word)` and `dbg(&item[0])` where you said, and the output showed the same: no space for the "Set" completion string, unlike the others.

Concerning your points on attaching images only if needed, how's this diff?
```diff
diff --git a/src/ui/popupmenu/completion_item_widget.rs b/src/ui/popupmenu/completion_item_widget.rs
index 95f31b5..e77de30 100644
--- a/src/ui/popupmenu/completion_item_widget.rs
+++ b/src/ui/popupmenu/completion_item_widget.rs
@@ -44,10 +44,10 @@ impl CompletionItemWidgetWrap {
         if show_kind {
             let buf = get_icon_pixbuf(&item.kind, icon_fg, size);
             image.set_from_pixbuf(&buf);
+            image.set_tooltip_text(format!("kind: '{}'", item.kind_raw).as_str());
+            image.set_margin_start(margin);
         }
 
-        image.set_tooltip_text(format!("kind: '{}'", item.kind_raw).as_str());
-        image.set_margin_start(margin);
         grid.attach(&image, 0, 0, 1, 1);
 
         let menu = gtk::Label::new(item.menu.as_str());
```

Attaching the images to the grid either way keeps the side padding when no icons are shown.

badosu: You forgot to move the `grid::attach` line, I just tested and it works for all unknown items:

Screenshot from 2019-06-19 07-43-11

badosu: It also works for matched items case:

Screenshot from 2019-06-19 07-47-13

The other concern indeed seems to be an issue with the LSP.

This is the whole working diff:

```diff
diff --git a/src/ui/popupmenu/completion_item_widget.rs b/src/ui/popupmenu/completion_item_widget.rs
index 95f31b5..a21beff 100644
--- a/src/ui/popupmenu/completion_item_widget.rs
+++ b/src/ui/popupmenu/completion_item_widget.rs
@@ -44,12 +44,11 @@ impl CompletionItemWidgetWrap {
         if show_kind {
             let buf = get_icon_pixbuf(&item.kind, icon_fg, size);
             image.set_from_pixbuf(&buf);
+            image.set_tooltip_text(format!("kind: '{}'", item.kind_raw).as_str());
+            image.set_margin_start(margin);
+            grid.attach(&image, 0, 0, 1, 1);
         }

-        image.set_tooltip_text(format!("kind: '{}'", item.kind_raw).as_str());
-        image.set_margin_start(margin);
-        grid.attach(&image, 0, 0, 1, 1);
-
         let menu = gtk::Label::new(item.menu.as_str());
         menu.set_halign(gtk::Align::End);
         menu.set_hexpand(true);
```

smolck: @badosu The reason I left the `grid.attach()` function was because of [this comment from vhakulinen]( https://github.com/vhakulinen/gnvim/pull/50#issuecomment-500590610):
> On a case when there are no icons shown (e.g. all kinds are unknown), remember to add some padding to the left side of the actual text.

Attaching the images to the grid keeps the padding on the left. Should we use less padding than when the icons are shown? Less padding may look better, but is it worth the extra logic? The normal padding amount may be too much when no icon is shown; what do you think? (See the bottom image of [this comment](https://github.com/vhakulinen/gnvim/pull/50#issuecomment-503220005) to see what it looks like with the normal padding)

badosu: Does this looks better?

Screenshot from 2019-06-19 14-07-38

I added to line 60:

```rust
if !show_kind {
  word.set_margin_start(5);
}
```

smolck: @badosu That definitely looks much better, good idea. Should I push the change, or wait to see what vhakulinen thinks?

vhakulinen: By some padding, I meant exactly what @badosu demonstrated.

smolck: Cool! I’ll push the change shortly.

badosu:
LGTM :+1:

vhakulinen:
Merged! Thanks for your efforts!

I made couple small changes: other was to remove unnecessary clone (effectively a copy) and added missing padding on info item when the icon is not shown.

Issue #53, opened 2019-06-11
Cursor Appears at Bottom of Screen When Moving Around File With Keyboard


smolck:
When moving in a file with the keyboard (`j`, `k`, `l`, `h`), the cursor first appears at the bottom right of the screen and then moves to where it should be in the file (This is a picture of it while the cursor is at the bottom right of the screen. After this, it appeared back at the correct position):

![image](https://user-images.githubusercontent.com/46855713/59311870-a7be4080-8c70-11e9-8796-45bf5aaa8907.png)

It does this erratically, and I have only noticed it while _holding down_ the keys to move (most notably when moving up or down in a file).

The following diff stops this behavior, but I imagine this is not an ideal solution and only masks the real problem (something with `neovim_lib`, maybe?):
```diff
diff --git a/src/ui/grid/grid.rs b/src/ui/grid/grid.rs
index b2e61b4..0af02df 100644
--- a/src/ui/grid/grid.rs
+++ b/src/ui/grid/grid.rs
@@ -57,7 +57,7 @@ pub struct Grid {
     /// EventBox to get mouse events for this grid.
     eb: EventBox,
     /// Internal context that is manipulated and used when handling events.
-    context: Arc<ThreadGuard<Option<Context>>>,
+    pub context: Arc<ThreadGuard<Option<Context>>>,
     /// Pointer position for dragging if we should call callback from
     /// `connect_motion_events_for_drag`.
     drag_position: Arc<ThreadGuard<(u64, u64)>>,
@@ -326,6 +326,11 @@ impl Grid {
         let mut ctx = self.context.borrow_mut();
         let ctx = ctx.as_mut().unwrap();
 
+        // Prevents cursor from appearing at bottom of screen
+        if row == ctx.rows.len() as u64 - 1 {
+            return;
+        }
+
         // Clear old cursor position.
         let (x, y, w, h) = ctx.get_cursor_rect();
         ctx.queue_draw_area
```

badosu:
I don't think this is an issue, I tried debugging it and noticed neovim's TUI presents the same behavior.

Seems to be an indicative of the current motion you're performing, e.g. `gg` gets you to the top of the page on normal mode, type the first `g` and notice it's shown there without flickering.

badosu:
It could be an issue if neovim provides an externalized api call to receive this data so that we can present it nicely.

smolck:
[...]

PR #54, opened 2019-06-14
[WIP] Add externalized messages support

badosu:
This is the base effort to start support for externalized messages.

My initial motivation was just to show regular messages (`"show_msg"` events), it seems that enabling this externalized setting changes drastically the UI though, passing all responsibility for anything resembling a message to it.

There are 4 kinds of separate UI components that are contemplated by ext_messages:

- Regular messages, these are sent by msg_show, cleared by msg_clear and shown in aggregate by msg_history_show. They span from messages outputted by plugins and vim to confirm dialogs, error messages, etc..
- Mode messages, these are shown when you are outside normal mode on the bottom of the editor when not externalized, "-- INSERT --", "-- VISUAL --", "-- VISUAL LINE --". Emmited by msg_showmode with content, cleared by same event without content.
- Cmd messages, the ones that show the current movement/command chord before conclusion, e.g. 'gg', 'ciw', etc... Emmited by msg_showcmd with content, cleared by same event without content.
- Ruler messages, current line and column being visualized, see :ruler. Emmited by msg_ruler with content, cleared by same event without content.

I would prefer an initial approach just externalizing regular messages and keeping everything else intact, will keep looking to see if that's possible.

Currently it only has the scaffold to deal with message events, it's a spike to see how that would work.

Using local neovim-lib until support for messages lands (https://github.com/daa84/neovim-lib/issues/32).

See: https://github.com/neovim/neovim/blob/master/runtime/doc/ui.txt#L680
A sample of events received: https://gist.github.com/badosu/80e3fe6adc232035cff8e99f3e798199

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ debug = true
 unstable = []
 
 [dependencies]
-neovim-lib = "0.6"
+neovim-lib = { git = "https://github.com/badosu/neovim-lib", branch = "badosu/add-ext-messages" }
 structopt = "0.2"
 
 gio = "0.5"
--- a/src/main.rs
+++ b/src/main.rs
@@ -129,6 +129,7 @@ fn build(app: &gtk::Application, opts: &Options) {
     ui_opts.set_cmdline_external(!opts.disable_ext_cmdline);
 
     ui_opts.set_wildmenu_external(true);
+    ui_opts.set_messages_external(true);
     nvim.ui_attach(80, 30, &ui_opts)
         .expect("Failed to attach UI");
 
--- a/src/nvim_bridge.rs
+++ b/src/nvim_bridge.rs
@@ -219,6 +219,12 @@ pub struct CompletionItem {
     pub info: String,
 }
 
+#[derive(Debug)]
+pub struct MessageEntry {
+    pub kind: String,
+    pub content: Vec<(u64, String)>,
+}
+
 pub struct PopupmenuShow {
     pub items: Vec<CompletionItem>,
     pub selected: i64,
@@ -280,6 +286,13 @@ pub enum RedrawEvent {
     WildmenuHide(),
     WildmenuSelect(i64),
 
+    MessageShow(MessageEntry, bool),
+    MessageShowMode(Vec<(u64, String)>),
+    MessageShowCmd(Vec<(u64, String)>),
+    MessageRuler(Vec<(u64, String)>),
+    MessageHistoryShow(Vec<MessageEntry>),
+    MessageClear(),
+
     Ignored(String),
     Unknown(String),
 }
@@ -324,6 +337,14 @@ impl fmt::Display for RedrawEvent {
             RedrawEvent::WildmenuShow(..) => write!(fmt, "WildmenuShow"),
             RedrawEvent::WildmenuHide(..) => write!(fmt, "WildmenuHide"),
             RedrawEvent::WildmenuSelect(..) => write!(fmt, "WildmenuSelect"),
+            RedrawEvent::MessageShow(..) => write!(fmt, "MessageShow"),
+            RedrawEvent::MessageShowMode(..) => write!(fmt, "MessageShowMode"),
+            RedrawEvent::MessageShowCmd(..) => write!(fmt, "MessageShowCmd"),
+            RedrawEvent::MessageClear(..) => write!(fmt, "MessageClear"),
+            RedrawEvent::MessageRuler(..) => write!(fmt, "MessageRuler"),
+            RedrawEvent::MessageHistoryShow(..) => {
+                write!(fmt, "MessageHistoryShow")
+            }
             RedrawEvent::Ignored(..) => write!(fmt, "Ignored"),
             RedrawEvent::Unknown(..) => write!(fmt, "Unknown"),
         }
@@ -489,6 +510,18 @@ GLOBALS:
     ["bell"]
     ["visual_bell"]
  */
+
+fn parse_hl_content(content: &Vec<Value>) -> Vec<(u64, String)> {
+    content
+        .into_iter()
+        .map(|args| {
+            let hl_id = unwrap_u64!(args[0]);
+            let text = unwrap_str!(args[1]).to_owned();
+
+            (hl_id, text)
+        })
+        .collect()
+}
 
 fn parse_redraw_event(args: Vec<Value>) -> Vec<RedrawEvent> {
     args.into_iter()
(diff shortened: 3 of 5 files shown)


src/main.rs
```diff
@@ -129,6 +129,7 @@ fn build(app: &gtk::Application, opts: &Options) {
     ui_opts.set_cmdline_external(!opts.disable_ext_cmdline);
 
     ui_opts.set_wildmenu_external(true);
+    ui_opts.set_messages_external(true);
```
badosu: Final version should have default false with option to enable (unless it gets really nice).

src/nvim_bridge.rs
```diff
@@ -282,6 +288,13 @@ pub enum RedrawEvent {
 
     Ignored(String),
     Unknown(String),
+
+    MessageShow(MessageEntry, bool),
+    MessageShowMode(Vec<(u64, String)>),
+    MessageShowCmd(Vec<(u64, String)>),
```
badosu: @smolck This is the event that shows those command indicators we were seeing on https://github.com/vhakulinen/gnvim/issues/53

Cargo.toml
```diff
@@ -10,7 +10,7 @@ debug = true
 unstable = []
 
 [dependencies]
-neovim-lib = "0.6"
+neovim-lib = { path = "../neovim-lib" }
```
badosu: Using local path until ext messages get supported on neovim-lib

src/nvim_bridge.rs
```diff
@@ -219,6 +219,12 @@ pub struct CompletionItem {
     pub info: String,
 }
 
+#[derive(Debug)]
+pub struct MessageEntry {
+    pub kind: String,
+    pub content: Vec<(u64, String)>,
```
badosu: This shows up often enough to be a good candidate for a struct, e.g. `HlContent`

smolck:
> I would prefer an initial approach just externalizing messages that use a dialog and keeping everything else intact, will keep looking to see if that's possible

So when you say dialog messages, you mean things like the output of wildmenu commands (`:!mkdir`, for example)? Forgive me for asking if this is obvious, but I want to clarify what is going on so I know before I try and help out.

badosu:
@smolck Anything that would be the result of `msg_show`: https://github.com/neovim/neovim/blob/master/runtime/doc/ui.txt#L693

There are 4 kinds of separate UI components that are contemplated by `ext_messages`:

- Regular messages, these are sent by `msg_show`, cleared by `msg_clear` and shown in aggregate by `msg_history_show`. They span from messages outputted by plugins and vim to confirm dialogs, error messages, etc..
- Mode messages, these are shown when you are outside normal mode on the bottom of the editor when not externalized, "-- INSERT --", "-- VISUAL --", "-- VISUAL LINE --"
- Cmd messages, the ones that show the current movement/command chord before conclusion, e.g. 'gg', 'ciw', etc...
- Ruler messages, current line and column being visualized, see `:ruler`.

Updated the description to make it more clear.

smolck:
Great, thanks for the clarification! That means once this is working, we could implement a nice external statusbar down the road, correct?

I tried compiling and running this setup (using your patched neovim-lib), but the application panics when trying to attach the Neovim UI, saying that ext_messages isn't a UI option. Any ideas why?

badosu:
> Great, thanks for the clarification! That means once this is working, we could implement a nice external statusbar down the road, correct?

Not exactly, the statusbar line is still the vim one, however many of the components present in the vicinity are externalized, see the sample I attached on the gist on the description.

> I tried compiling and running this setup (using your patched neovim-lib), but the application panics when trying to attach the Neovim UI, saying that ext_messages isn't a UI option. Any ideas why?

Currently `neovim-lib` does not support it so I [sent a PR](https://github.com/vhakulinen/gnvim/pull/54) to add it. I updated the current PR to use the branch.

Cargo.toml
```diff
@@ -10,7 +10,7 @@ debug = true
 unstable = []
 
 [dependencies]
-neovim-lib = "0.6"
+neovim-lib = { git = "https://github.com/badosu/neovim-lib", branch = "badosu/add-ext-messages" }
```
badosu: Using custom branch until support lands.

smolck: Interesting. Pulled this updated PR, and all I am getting on running it is this error:
`thread ‘main’ panicked at ‘Failed to attach UI: NeovimError(1, “No such option: ext_messages”)’`

This shouldn’t be happening, since I am using your neovim-lib that you updated to support the messages?

badosu: Are you using a recent neovim version?

smolck: Yup, version 0.3.7

badosu: [I think](https://github.com/neovim/neovim/pull/7466) this is only available on nvim 0.4.0 which is unreleased. Depending on your OS you can easily install the nightly, or have to build it yourself.

badosu: Again, this is just a proof-of-concept, not something to be too concerned about at this stage.

vhakulinen:
My efforts for ext_messages support from couple weeks ago: https://github.com/vhakulinen/gnvim/compare/ext_messages?expand=1. I got distracted by the removal of ext_wildmenu which got incorporated with ext_popupmenu in https://github.com/neovim/neovim/pull/9607.

src/nvim_bridge.rs
vhakulinen: `type HlContent = Vec<u64, String>` or `type HlContent = Vec<HlText>; type HlText = (u64, String>` would be better than a struct with a single property.

vhakulinen:
> > Great, thanks for the clarification! That means once this is working, we could implement a nice external statusbar down the road, correct?
>
> Not exactly, the statusbar line is still the vim one, however many of the components present in the vicinity are externalized, see the sample I attached on the gist on the description.

You can use `set laststatus=0` to get rid of the vim's status line.

vhakulinen:
Also be aware of this: https://github.com/neovim/neovim/pull/10065

badosu:
> My efforts for ext_messages support from couple weeks ago: https://github.com/vhakulinen/gnvim/compare/ext_messages?expand=1

Oh, I was not aware of that!

vhakulinen:
Closing in favor of #57. If you want to put any development efforts in implementing ext messages support, please base the work off #57. I really much want to make ext messages support a thing in gnvim but my time is limited so any help is appreciated (as always!). We should gather any experience/opinions/etc. on #57.

badosu:
:+1:

PR #56, opened 2019-06-16
Wildmenu -> pmenu events MVP

vhakulinen:
Just throwing this here for now.


smolck:
Should this be merged before we work on #57? (And if so, what further work needs to be done here, if any?)

vhakulinen:
Maybe. This requires nvim 0.4.0 which is not released yet, while ext_messages API exists on version 0.3.

bfredl:
`ext_messages` is on master only and not in 0.3.x

vhakulinen:
Oh, that's right.

Issue #38, opened 2019-04-30
Indentation issue in Terminal


badosu:
### How to reproduce

- Open a `:terminal`
- Run `ls -al`

### What is expected

Output is correctly indented

### What happens

![Screenshot from 2019-04-30 15-47-13](https://user-images.githubusercontent.com/347552/56985670-41b5a800-6b5f-11e9-9333-dae16fb69c3e.png)

It works correctly on [gonvim](https://github.com/akiyosi/gonvim), so I don't think it's an externalized issue.

vhakulinen:
Try running `reset` before `ls -as`. That looks more like a problem with the terminal it self than with gnvim.

badosu:
Yeah, it works!

Do you mind sharing why reset fixes this?

vhakulinen:
It has something to do with a think called terminal mode (if that is its name). `ncurses` and other TUIs (like nvim) changes the terminal mode settings which upon an unclean exit causes the terminal mode to be "invalid" when the terminal returns into the command prompt. `reset` command resets the terminal mode.

badosu:
Thanks for the info @vhakulinen, currently I am prepending `reset` to my commands but I'll investigate why I have this discrepancy present, why it happens on gnvim but not on nvim or gonvim.

smolck:
@vhakulinen Do you get this issue (and need to run `reset`) when you use `:terminal` in GNvim? If so, should we reopen this? I get this issue as well (even with [Nuake](https://github.com/Lenovsky/nuake)), and will hopefully be able to look into a possible fix (though, that may not be necessary; if you don't get this issue, I'm not sure if we should look into fixing it).

vhakulinen:
This is not a problem with gnvim, but with the shell in use and/or nvim terminal. What ever is causing this issue in gnvim, will also occur with TUI if executed in the _exact_ same environment (same env vars and so on).

I've noticed that some prompts (`PS1`) causes this issue.

badosu:
This does not happen for me in neovim.

PR #57, opened 2019-06-23
Implement ext messages

vhakulinen:
Throwing this here for now. It works, but this is not ready to be merged yet. Feel free to test, play around, and do further development on this. Feedback is welcome.


smolck:
@vhakulinen What are the main features/changes that you want to see in this PR for it to be ready for merging?

vhakulinen:
At least support for `msg_show` and `msg_history_show`. For `msg_show`, it would be nice to have some "special" support for the confirm/prompt messages. `msg_showmode`, `msg_showcmd`, and `msg_ruler` would be most useful within a custom statusbar, so those would be a matter of another PR.

smolck:
@badosu This PR fixes that indentation issue from #38 for me; if you get the chance, maybe try this branch out and see if the same is true for you. (With this branch I don't need to use `reset`)

src/ui/messages.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    pub fn show(&mut self, msg: &MsgShow, hl_defs: &HlDefs) {
+
+        if msg.replace_last {
+            self.messages.pop();
+        }
+
+        let msg = Message::new(msg, hl_defs, &self.css_provider, self.font.height as f64);
+        self.container.pack_end(&msg.widget(), false, true, 5);
```
smolck: Is the last parameter of `self.container.pack_end` here meant to be `5`? When set to 5, there is a small bit of padding between the GNvim window and the message (notice the gap between the top of the message and the window):
image

With `0` used instead, it looks like this, with no gap between the message and window, which I think is a bit cleaner:
image

If we wanted to keep the padding, we might want to add some to the right too (although that might look strange, haven't tried it).

smolck:
I now have a working implementation of support for `msg_history_show` messages. I've added some explanatory comments (unnecessary in the actual commit) of some of the things I've done. I'm sure there is a better way of unwrapping the messages (as I mentioned in the comments) than what I did, but this is just a draft. Also, there is no 'press ENTER' prompt to remove the messages as there is in the TUI (we probably want that in the final version?). Let me know what you think (I've put it into a gist since it's fairly large; some of the changes are only from `cargo fmt`):

[Draft commit for msg_history_show message support](https://gist.github.com/smolck/171a33fbcc286cbd3db04bf66f55792d)

badosu:
@smolck 

> Also, there is no 'press ENTER' prompt to remove the messages as there is in the TUI

Look at 'confirm', 'confirm_sub' and 'return_prompt' messages on [ui.txt](https://github.com/neovim/neovim/blob/master/runtime/doc/ui.txt#L697-L710)

vhakulinen:
> Also, there is no 'press ENTER' prompt to remove the messages as there is in the TUI

Also see https://github.com/neovim/neovim/pull/10065.

src/ui/messages.rs
vhakulinen: I added the padding there so there would indeed be some spacing between the messages.

smolck:
> It's not clear how to clear 'msg_history_show' though, I'd assume it's 'msg_clear' but in the docs it says that it clears only messages shown by 'msg_show'.

Interesting. That's right; I did a few tests, and `msg_clear` isn't sent after `msg_history_show`. `msg_history_show` is only cleared after the next message receives the `msg_clear` command (in gnvim, that is; in the TUI there is a 'press ENTER' prompt to clear it).

smolck:
@vhakulinen Will you be able to submit a PR over at [neovim-lib](https://github.com/daa84/neovim-lib) for supporting multi grid (when you get the time), that way there will be no need to use a personal fork? It already has ext_message support from [badosu’s PR](https://github.com/daa84/neovim-lib/pull/33), so multi grid support is all that’s needed.

badosu:
@smolck As far as I can see that should not be necessary at the moment, just using the upstream master branch should do it (this PR does not make use of multigrid).

bfredl:
> It's not clear how to clear 'msg_history_show' though, I'd assume it's 'msg_clear' but in the docs it says that it clears only messages shown by 'msg_show'.

Currently nvim delegates all choices how to navigate the message history (if it doesn't fit into one screen), how to dismiss it, etc to the client. Somehow running the builtin pager over ext_messages seems hard, nvim doesn't know how much space messages take on the client's screen.

> Also see neovim/neovim#10065.

This is a merged PR. Is there any remaining issue with it?

smolck:
> Currently nvim delegates all choices how to navigate the message history (if it doesn't fit into one screen), how to dismiss it, etc to the client.

So we'll need to decide how we want to handle those things. What about having the message history be a scrolled window? When it doesn't fit into one screen, the user could scroll through it, and if he doesn't interact with it, we could make it fade away after a few seconds. I think we'd need to create a new struct for the purpose though (it seems that might be the best option for showing the messages from `msg_history_show` anyway; it gives us more options).

vhakulinen:
> This is a merged PR. Is there any remaining issue with it?

I was trying to point out that you need to have recent master of nvim with this. Tho' you need quite recent version of it either way to have ext_messages support.

vhakulinen:
> What about having the message history be a scrolled window?

Thats probably the way to go. I'm not sure about the UX and here I would try to see what other GUIs do. 

Opening a split to the bottom might be a decent solution too since many editors (afaik) does so, like so:

image

That might be a bit confusing regarding vim's other features, such as the quickfix window.

badosu:
Just a remark: I think that the way messages are displayed in neovim are more 'log-like' than 'notification-like'.

I think a pretty straightforward implementation would be a console-like component below the statusbar that displays messages as line, with some keybindings/UI/behavior (closing, hiding, confirming, etc). This also helps with less distraction potential (which I think is one of the best features of gnvim).

I am not saying the current implementation should adhere to it, just my two cents on this topic.

smolck:
> Just a remark: I think the way messages are displayed in neovim are more 'log-like' than 'notification-like'.

Agreed.

> I think a pretty straightforward implementation would be a console-like component below the statusbar that displays messages as line . . .

When you say 'displays messages as line', do you mean like what vhakulinen showed above?

badosu:
> Of course, if the user didn't want any of this he could always disable ext_messages with command args to gnvim

This is somewhat awkward to set at the moment, requires fiddling with `.desktop` file or running as a command, in the future we might want to improve this. I created https://github.com/vhakulinen/gnvim/issues/59 to follow up on it.

vhakulinen:
Experience report: especially long messages gets easily on way of the cursor, which is _really_ annoying. The current solution needs to move them messages up/down according to the cursor position _or_ make the messages narrower (I get full screen wide messages from lsp). Other solution would be to not display the messages on top of the grid. Or some middle ground would work too: some messages are displayed as they are currently in this branch and some on the bottom in its own panel which pushes the grid upwards. Dunno.

vhakulinen:
Experiences with `ext_messages` will vary very much based on what plugins one has enabled.

smolck:
@vhakulinen We might want to check how ext_messages are supported in [this branch of gonvim (ext_messages)](https://github.com/akiyosi/gonvim/tree/ext_messages), and/or [this one (0.4.0)](https://github.com/akiyosi/gonvim/tree/0.4.0). They might not have support like we’re trying to implement (I haven’t looked into them), but maybe still worth a look to get some ideas.

smolck:
> The current solution needs to move them messages up/down according to the cursor position _or_ make the messages narrower (I get full screen wide messages from lsp). 

Moving the messages up/down might be a bit distracting, and making the messages narrower lengthwise would make them less readable. Making them narrower vertically-speaking would only work up until a certain font size, right?

> Other solution would be to not display the messages on top of the grid. Or some middle ground would work too: some messages are displayed as they are currently in this branch and some on the bottom in its own panel which pushes the grid upwards.

It seems keeping the/some messages as they currently are implemented may require unnecessary boilerplate. How about making _all_ (or most) messages from `msg_show` be at the bottom (or top, side, etc., make it user-configurable) in their own panel that moves the grid accordingly, and that is shown/hidden with a (user-configurable) keybinding? If the panel was hidden by the user through a keybinding, we could show the messages like the TUI does at the bottom of the screen. What would be nice about this is that it would be consistent with how we decided we want to do the `msg_history_show` support (using a panel at the bottom like other text editors). `msg_history_show` might not even be necessary, as we could keep all of the messages listed in the panel, until they are cleared by the user or too many are shown ('too many' would be a user-configurable number of messages). We may still want a dialog for confirm/prompt messages though. (Let me know if I wasn't clear on any of this.)

smolck:
@vhakulinen @badosu So, are we decided on how we want to go about implementing this? I’d like to help out where I can, so when you get the chance let me know what we to do, if anything (assuming we’re decided on this).

vhakulinen:
I've not done much with this in a while, and am now focusing my efforts to #95. Own panel/container for the messages on the bottom is probably the best solution. If you want to work on this, you could start by doing that.

vhakulinen:
Which ever you prefer is fine. There is some base work in this branch which is likely to benefit you, so this branch might be better option.

vhakulinen:
Rebased (I actually had done this a while back on my local already). Made the neovim-lib dependency to point to the upstream one.

PR #58, opened 2019-06-24
Enhance README.md

smolck:
I have added the GNvim logo and a screenshot from the wiki to the README, in addition to changing a few small things. Let me know what you think; personally, I think the logo and screenshot make the project more appealing. (The screenshot I used doesn't necessarily need to be the final one, I just didn't want to deviate from what was already in the wiki. And that's assuming this change is wanted.)


vhakulinen:
I like it! Might want to update the screenshot at some point.

Issue #59, opened 2019-06-25
Improve configurability


badosu:
At the moment, configuration requires changing the `.desktop` file or running gnvim as a command.

It would be ideal to keep as much configuration as possible as vim global variables or respecting vim options, but if not feasible (e.g. for externalized features) use a xdg config file.

smolck:
I haven't looked into this, but [neovim-gtk has some decent-looking configuration](https://github.com/daa84/neovim-gtk/wiki/Configuration) (see [here](https://github.com/daa84/neovim-gtk/wiki/Configuration#options) for the way it handles disabling external features). Not sure if we want to go the same route. It might be nice to have a simple option of `g:gnvim_external_tabline` or something that could be set to 0 or 1 in `init.vim` or similar (although we might want something different for on-the-fly configuring).

vhakulinen:
Looks like `neovim-gtk` is just calling [set_option](https://github.com/daa84/neovim-gtk/blob/e4482a679963177d2affb91131636e872dfe6c14/src/nvim/redraw_handler.rs#L150), so doing the same on gnvim shouldn't be too hard.

smolck:
How do we want to go about exposing the configuration options to the user (instead of command line args)? Once I know I can probably start working on a PR for this (I imagine most if not all of the refactoring will go into `main.rs`?).

vhakulinen:
Messages over rpc might be the only way to do this currently. Thats how the ["gnvim" events](https://github.com/vhakulinen/gnvim/blob/5bf01b9b965d580355b7abc66419dee1c93d1c58/src/nvim_bridge/mod.rs#L909) are done currently.

smolck:
@vhakulinen @badosu Okay, so with that info I have a (fairly small!) working diff, let me know if this is worth opening a PR for (right now you have to manually `:call rpcnotify(g:gnvim_channel_id, 'Gnvim', 'DisableExtTabline')`; I assume we'll want to create a commands.vim file or similar for it):
```diff
diff --git a/src/nvim_bridge/mod.rs b/src/nvim_bridge/mod.rs
index 2a723fc..06e2cd8 100644
--- a/src/nvim_bridge/mod.rs
+++ b/src/nvim_bridge/mod.rs
@@ -424,6 +424,7 @@ impl fmt::Display for RedrawEvent {
 #[derive(Debug, PartialEq)]
 pub enum GnvimEvent {
     SetGuiColors(SetGuiColors),
+    DisableExtTabline,
[...]

PR #61, opened 2019-07-02
Improve configurability with new rpc messages

smolck:
This mostly works, with only one current problem: when the user disables the ext_tabline while it is in use, it doesn't disappear (and couldn't be made to go away through closing tabs when I tested), so that he ends up with something like this (gnvim's tabs won't go away):

image

Also, after disabling the ext_tabline the TUI tabline doesn't appear until the user changes tabs (at least when I tested it). I can work on fixing this once we know how we want to go about doing so (I don't want to bloat the code base with a bad solution). (As an FYI I am using Neovim master from about a week ago, give or take.) I don't think this could be a problem with the popupmenu, since one can't call the commands to enable/disable it without exiting insert mode (may be wrong though).

If we want to add any other configuration options, let me know and I'll add them. Any feedback is much appreciated. (Note that once things are solidified here, we’ll probably want to add documentation for the new commands. Don’t want to do that just yet though, since we may want to change the command names, add more commands, etc.)

See #59.

### DIFF
--- a/src/nvim_bridge/mod.rs
+++ b/src/nvim_bridge/mod.rs
@@ -818,6 +818,10 @@ impl fmt::Display for RedrawEvent {
 #[derive(Debug, PartialEq)]
 pub enum GnvimEvent {
     SetGuiColors(SetGuiColors),
+    EnableExtTabline(bool),
+    EnableExtCmdline(bool),
+    EnableExtPmenu(bool),
+
     CompletionMenuToggleInfo,
 
     CursorTooltipLoadStyle(String),
@@ -1071,6 +1075,42 @@ pub(crate) fn parse_gnvim_event(
 ) -> Result<GnvimEvent, String> {
     let cmd = try_str!(args.get(0).ok_or("No command given")?, "cmd");
     let res = match cmd {
+        "EnableExtCmdline" => {
+            let option = try_str!(
+                args.get(1).ok_or("No input for EnableExtCmdline")?,
+                "enable ext cmdline"
+            )
+            .parse();
+            if let Err(error) = option {
+                return Err(format!("Didn't receive a boolean: {}", error));
+            }
+
+            GnvimEvent::EnableExtCmdline(option.unwrap())
+        }
+        "EnableExtTabline" => {
+            let option = try_str!(
+                args.get(1).ok_or("No input for EnableExtTabline")?,
+                "enable ext tabline"
+            )
+            .parse();
+            if let Err(error) = option {
+                return Err(format!("Didn't receive a boolean: {}", error));
+            }
+
+            GnvimEvent::EnableExtTabline(option.unwrap())
+        }
+        "EnableExtPmenu" => {
+            let option = try_str!(
+                args.get(1).ok_or("No input for EnableExtPmenu")?,
+                "enable ext pmenu"
+            )
+            .parse();
+            if let Err(error) = option {
+                return Err(format!("Didn't receive a boolean: {}", error));
+            }
+
+            GnvimEvent::EnableExtPmenu(option.unwrap())
+        }
         "SetGuiColors" => {
             let mut colors = SetGuiColors::default();
 
--- a/src/ui/ui.rs
+++ b/src/ui/ui.rs
@@ -446,6 +446,35 @@ fn handle_gnvim_event(
                         err
                     ));
                 });
+
+            if !*option {
+                state.tabline.get_widget().hide();
+                return;
+            }
+
+            match (nvim.get_current_tabpage(), nvim.list_tabpages()) {
+                (Ok(current), Ok(tabpages)) => {
+                    let tabpages = tabpages
+                        .iter()
+                        .map(|x| (x.clone(), String::from("")))
+                        .collect();
+
+                    state.tabline.update(&mut nvim, current, tabpages);
+                }
+                (Err(err), Ok(_)) => {
+                    nvim.command_async(&format!(
+                            "echo \"Unable to get current tabpage for updating tabline: '{}'\"", err));
+                }
+                (Ok(_), Err(err)) => {
+                    nvim.command_async(&format!(
+                            "echo \"Unable to get tabpages for updating tabline: '{}'\"", err));
+                }
+                (Err(get_current_err), Err(list_tabpages_err)) => {
+                    nvim.command_async(&format!(
+                            "echo \"Unable to get current tabpage for updating tabline: '{}'\n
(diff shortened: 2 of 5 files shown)


smolck:
So, this diff mostly fixes the problem with the tabline not going away when it is disabled (it is at the bottom of the `GnvimEvent::DisableExtTabline` match arm):
```diff
diff --git a/src/ui/ui.rs b/src/ui/ui.rs
index 9c76c5b..fc31586 100644
--- a/src/ui/ui.rs
+++ b/src/ui/ui.rs
@@ -445,6 +445,7 @@ fn handle_gnvim_event(
                     ));
                 },
             );
+            state.tabline.get_widget().hide();
         }
         GnvimEvent::DisableExtPopupmenu => {
             let mut nvim = nvim.lock().unwrap();
```

Only problem left is that if buffers are shown in the TUI tabline, which is possible with vim-airline, the area where the TUI tabline previously was shown is blank until the window is updated (i.e. by changing tabs, creating a split, etc., even resizing the window I think) when the external tabline is re-enabled.

badosu:
Perhaps sending a `RedrawEvent::TablineUpdate` fixes this?

smolck:
How should I send that `RedrawEvent`? I can't really call `handle_redraw_event` directly with the current setup (and if I could I'd need the tabpages and the strings associated with them).

badosu:
Hmm, I thought exactly about that method, we should at least have this available for these cases so might be necessary to build it if not accessible now.

smolck:
@badosu That should work; I’ll give it a try. There’s probably API methods for getting the info we want, so this may be pretty simple. I’ll post a diff if I get one working.

smolck:
The problem I'm now running into is getting the name of each tab. That information is only gotten from [here](https://github.com/vhakulinen/gnvim/blob/master/src/nvim_bridge/mod.rs#L811) AFAIK. Of course, I could always find a way to pass that information around, but that feels unnecessary. I can get the tabpages and the current tabpage via API methods, but the name of the Tabpage is not easily accessible ATM. I may be missing something though; any ideas?

kenardes:
hello, noobs here, how to show / enable vim-airline tabline? it's shown on TUI / terminal, but not in gnvim. I need it to easily see buffer list.

smolck:
@kenardes Until this PR is merged, try running GNvim like this:

`$ gnvim --disable-ext-tabline`

If that doesn't work or I misunderstood, please let me know.

kenardes:
> If that doesn't work or I misunderstood, please let me know.

it work, thanks.
but is there away to apply same option "disable-ext-tabline" in vim config file ?

smolck:
> but is there away to apply same option "disable-ext-tabline" in vim config file ?

At the moment, master has no way of doing that. However, if you're willing to use this branch (note that it isn't done yet), then you can put this in your `.vimrc`/`init.vim`:

```vim
if exists('g:gnvim')
    call gnvim#tabline#disable()
endif
```

But again, the above will only work with this branch.

@vhakulinen Do you mind giving this branch a review once you get the chance? I know that documentation needs to be taken care of, but is there anything else?

vhakulinen:
Only note so far: remove the `Disable...` messages, and make the `Enable...` messages to take a boolean argument which indicates if given feature should be enabled or not. The cli arguments can stay (for now at least).

P.S. I'm going to start my holidays in a week, so hopefully I find the energy to invest my time into gnvim again.

smolck:
@vhakulinen @badosu Besides the conflicts and maybe some other small things, the main problem with this branch right now is the problem with the tabline mentioned above. @badosu suggested manually issuing a redraw event for the tabline; is that the direction we want to go? If so, we’ll need a way to get the necessary info for issuing the redraw event (getting that info’s where I got stuck).

smolck:
This branch is now ready for testing and review. I fixed the problems with the tabline, using @badosu's suggestion of getting the information for the tabline and manually updating it. Let me know if you spot any issues, inconsistencies, etc.

cc: @vhakulinen

src/nvim_bridge/mod.rs
```diff
@@ -820,6 +820,10 @@ impl fmt::Display for RedrawEvent {
 #[derive(Debug, PartialEq)]
 pub enum GnvimEvent {
     SetGuiColors(SetGuiColors),
+    EnableExtTabline(bool),
+    EnableExtCmdline(bool),
+    EnableExtPmenu(bool),
```
smolck: Should these say `DisplayExt ...` instead of `EnableExt ...`?

src/ui/ui.rs
```diff
@@ -408,6 +408,66 @@ fn handle_gnvim_event(
...
+                        err
+                    ));
+                });
+
+            if !*option {
+                state.tabline.get_widget().hide();
+                return;
+            }
+
+            match (nvim.get_current_tabpage(), nvim.list_tabpages()) {
```
smolck: This error handling may be excessive, if so let me know.

src/ui/ui.rs
```diff
@@ -408,6 +408,66 @@ fn handle_gnvim_event(
...
+                    let tabpages = tabpages
+                        .iter()
+                        .map(|x| (x.clone(), String::from("")))
+                        .collect();
+
+                    state.tabline.update(&mut nvim, current, tabpages);
+                }
+                (Err(err), Ok(_)) => {
+                    nvim.command_async(&format!(
+                            "echo \"Unable to get current tabpage for updating tabline: '{}'\"", err));
```
smolck: Better error msg?

src/ui/ui.rs
```diff
@@ -408,6 +408,66 @@ fn handle_gnvim_event(
...
+                }
+                (Ok(_), Err(err)) => {
+                    nvim.command_async(&format!(
+                            "echo \"Unable to get tabpages for updating tabline: '{}'\"", err));
```
smolck: Same as above.

src/ui/ui.rs
```diff
@@ -408,6 +408,66 @@ fn handle_gnvim_event(
...
+                }
+                (Err(get_current_err), Err(list_tabpages_err)) => {
+                    nvim.command_async(&format!(
+                            "echo \"Unable to get current tabpage for updating tabline: '{}'\n
+                            Unable to get tabpages for updating tabline: '{}'\"", get_current_err, list_tabpages_err));
```
smolck: Same as above.

Famiu:
Why is this not merged yet? Currently it's a pain to disable GNvim features, I feel like this should be merged asap.

Issue #65, opened 2019-07-09
guifont: floating-point is not supported?


last-partizan:
Hello.

Looks like floating-point numbers is not supported if font definition?

I trying to set `guifont=Monospace:h11.25`, but with no effect.
Or maybe there is another way?

badosu:
I've never seen this use case, is this because of HiDPI display? In which case perhaps we could also fix with better support?

last-partizan:
No, this is not HiDPI, this is some strange combination of freetype bugs (or features) and font (i'm using https://github.com/source-foundry/Hack, but it was the same with DejaVu Sans Mono)

When font-size is set to 11, somehow font height is smaller than it must be and it looks weird. When i set size to 11.25, font looks like expected.

here is what it looks like
![Screenshot from 2019-07-09 14-40-38](https://user-images.githubusercontent.com/301015/60885330-05d24980-a258-11e9-99d0-b05ba3b497d1.png)
![Screenshot from 2019-07-09 14-40-10](https://user-images.githubusercontent.com/301015/60885332-05d24980-a258-11e9-9a90-f26b01c84432.png)

smolck:
@badosu I think the problem is in font.rs [here](https://github.com/vhakulinen/gnvim/blob/master/src/ui/font.rs#L50). It tries to parse the font to type `usize`, which doesn’t support floating-point numbers and thus fails. A fairly easy fix, I think: just swap the font height in the `Font` struct to an `f32` or similar and parse it to that floating-point type instead (and make other changes elsewhere to accommodate that change accordingly). If we want I can make a PR for it, though @vhakulinen might just want to do the commit, since I think it’d be fairly small. (All this is assuming that `f32` would work, haven’t tested it/investigated. Let me know if I’m wrong on any of this.)

badosu:
@smolck Like the `WM_CLASS` commit I think it's better to send PRs. It gives a single point of discussion and I can simply go grab your fork and test it.

smolck:
@last-partizan As of #66, GNvim now supports floating-point guifont height.

PR #66, opened 2019-07-09
Add support for floating-point guifont height

smolck:
This works as intended for me, let me know what you think.

See #65.

### DIFF
--- a/src/ui/font.rs
+++ b/src/ui/font.rs
@@ -1,7 +1,7 @@
 use std::fmt;
 use std::fmt::Display;
 
-const DEFAULT_HEIGHT: usize = 14;
+const DEFAULT_HEIGHT: f32 = 14.0;
 
 pub enum FontUnit {
     Pixel,
@@ -20,7 +20,7 @@ impl Display for FontUnit {
 #[derive(Clone)]
 pub struct Font {
     name: String,
-    pub height: usize,
+    pub height: f32,
 }
 
 impl Font {
@@ -47,8 +47,8 @@ impl Font {
                 match ch {
                     'h' => {
                         let rest = chars.collect::<String>();
-                        let h = rest.parse::<usize>().or(Err(()))?;
-                        if h == 0 {
+                        let h = rest.parse::<f32>().or(Err(()))?;
+                        if h <= 0.0 {
                             // Ignore zero sized font.
                             continue;
                         }
@@ -114,7 +114,7 @@ mod tests {
     fn test_as_wild_css() {
         let font = Font {
             name: "foo".to_string(),
-            height: 10,
+            height: 10.0,
         };
 
         assert_eq!(
@@ -139,14 +139,12 @@ mod tests {
         // Font with proper height.
         let f = Font::from_guifont("monospace:h11").unwrap();
         assert_eq!(f.name, "monospace");
-        assert_eq!(f.height, 11);
+        assert_eq!(f.height, 11.0);
 
         // Font with invalid height.
         let f = Font::from_guifont("font:h");
         assert_eq!(f.is_err(), true);
         let f = Font::from_guifont("font:hn");
-        assert_eq!(f.is_err(), true);
-        let f = Font::from_guifont("font:h-1");
         assert_eq!(f.is_err(), true);
 
         // Font with height zero.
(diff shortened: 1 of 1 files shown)


badosu:
Works for me!

Some tests are failing though, current one because previously font size was unsigned and float accepts negative values.

I think it's acceptable to ignore negative values as we are doing right now, but if previous behavior is wanted you can split the condition on https://github.com/smolck/gnvim/blob/9ea48e92fab0730b1530ce4145fb76be49332635/src/ui/font.rs#L51 and return `Err()` when negative.

last-partizan:
I just compiled it from your branch and it works like expected. Thanks!

One more point, how do i set guifont in config? does gnvim have something like `.gnvimrc`? or `has('gui_gnvim`)`?

smolck:
@last-partizan This should work, since GNvim sets `g:gnvim` to 1 when it is running:

```
if exists(‘g:gnvim’)
    set guifont=<your font and height>
    “ Other settings for GNvim
endif
```

This isn’t strictly necessary, since doing a simple `set guifont=...` in your `init.vim`/`.vimrc` shouldn’t break anything and would work as you want (I think).

smolck:
@badosu Nice spot! Didn't notice that. How about this?
```diff
diff --git a/src/ui/font.rs b/src/ui/font.rs
index 4d4ecf2..9bc5cd6 100644
--- a/src/ui/font.rs
+++ b/src/ui/font.rs
@@ -48,6 +48,7 @@ impl Font {
                     'h' => {
                         let rest = chars.collect::<String>();
                         let h = rest.parse::<f32>().or(Err(()))?;
+
                         if h == 0.0 || h < 0.0 {
                             // Ignore zero sized font.
                             continue;
@@ -146,15 +147,18 @@ mod tests {
         assert_eq!(f.is_err(), true);
         let f = Font::from_guifont("font:hn");
         assert_eq!(f.is_err(), true);
-        let f = Font::from_guifont("font:h-1");
-        assert_eq!(f.is_err(), true);
 
         // Font with height zero.
         let f = Font::from_guifont("foo:h0").unwrap();
         assert_eq!(f.name, "foo");
         assert_eq!(f.height, DEFAULT_HEIGHT);
 
-        // Font with no hegith.
+        // Font with negative height.
+        let f = Font::from_guifont("font:h-1").unwrap();
+        assert_eq!(f.name, "font");
+        assert_eq!(f.height, DEFAULT_HEIGHT);
+
+        // Font with no height.
         let f = Font::from_guifont("bar").unwrap();
         assert_eq!(f.name, "bar");
         assert_eq!(f.height, DEFAULT_HEIGHT);
```

I think ignoring the negative font is a better option. If someone types a negative number for height, I think changing their font type might be annoying. It'd be nice to set the font size to default to the current font size (if it is set) when an invalid font size is given, but that's a matter for another PR (and a different file).

badosu:
> I think ignoring the case of a negative font is better than returning an error

I agree with this approach :+1:.

src/ui/font.rs
```diff
@@ -139,22 +139,25 @@ mod tests {
...
-        assert_eq!(f.is_err(), true);
 
         // Font with height zero.
         let f = Font::from_guifont("foo:h0").unwrap();
         assert_eq!(f.name, "foo");
         assert_eq!(f.height, DEFAULT_HEIGHT);
 
-        // Font with no hegith.
+        // Font with negative height.
+        let f = Font::from_guifont("font:h-1");
```
badosu:
```suggestion
         let f = Font::from_guifont("font:h-1").unwrap()
```

smolck: Thanks @badosu! Didn't see that!

src/ui/font.rs
```diff
@@ -47,8 +47,8 @@ impl Font {
                 match ch {
                     'h' => {
                         let rest = chars.collect::<String>();
-                        let h = rest.parse::<usize>().or(Err(()))?;
-                        if h == 0 {
+                        let h = rest.parse::<f32>().or(Err(()))?;
+                        if h == 0.0 || h < 0.0 {
```
vhakulinen: `h <= 0.0` is enough.

smolck: @vhakulinen Thank you! Should be fixed now.

PR #67, opened 2019-07-10
[WIP] Only print debug messages when compiled in debug mode

smolck:
This is an idea of how we could make GNvim print debug statements only when compiled with `cargo run` (i.e. in debug mode). It isn't ready for merging yet, as it needs more `#[cfg(debug_assertions)]` statements and needs to silence the generated warnings (`variable_unused` warnings due to the code not always being there. A simple workaround would be to prefix the variables with underscores). That’s assuming we want to do this, and do it in this way.

The reason for this comes from users’ seeing the error/println messages when running gnvim in release mode (i.e. when it is installed).

This could make the code base less readable, so I am not sure if we want this. But I figured I'd throw it out there anyways, just to show it is an option. (Of course we could just remove the debug `println!` statements, but this is a way we can keep them in the code.) Other approaches are probably possible, and if we continue forward we may want to look into them.

Any feedback is much appreciated.

See:
https://doc.rust-lang.org/beta/reference/conditional-compilation.html
https://users.rust-lang.org/t/conditional-compilation-for-debug-release/1098/8
`debug_assertions`: https://doc.rust-lang.org/cargo/reference/manifest.html#the-profile-sections
debug/release conditional compilation may also be improved in the future: https://github.com/rust-lang/rust/issues/59999

### DIFF
--- a/src/ui/ui.rs
+++ b/src/ui/ui.rs
@@ -421,7 +421,8 @@ fn handle_notify(
                 ))
                 .cb(|res| match res {
                     Ok(_) => {}
-                    Err(err) => {
+                    Err(err) =>
+                    {
                         #[cfg(debug_assertions)]
                         println!("Failed to execute nvim command: {}", err)
                     }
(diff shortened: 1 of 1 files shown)


smolck:
There is no need for this, I don't think. Should have thought more about this before creating a PR; apologies, my mistake. Closing.

PR #68, opened 2019-07-10
Fix a few minor issues (compiler warnings)

smolck:
I noticed that the compiler gave me some warnings when compiling GNvim (with cargo nightly), so I figured I'd fix them (you can see them in the [cargo nightly Travis checks](https://travis-ci.org/vhakulinen/gnvim/jobs/555885864)). Let me know if there's any reason these changes won't work (like if they could/do cause other bugs, mess with compatibility, etc.), or if you have any other feedback.

As a side note, I've had @badosu's issue he mentioned in #38, and I noticed that this branch fixes it for me. Not sure why that is?

### DIFF
--- a/src/ui/popupmenu/lazy_loader.rs
+++ b/src/ui/popupmenu/lazy_loader.rs
@@ -19,7 +19,8 @@ struct State {
     /// Once we're loaded some (or all) data, this closure gets called if
     /// one exists. The first value in the tuple can is indication on the
     /// number of items needed before calling the closure.
-    once_loaded: Option<(Option<i32>, Box<Fn(&Vec<CompletionItemWidgetWrap>)>)>,
+    once_loaded:
+        Option<(Option<i32>, Box<dyn Fn(&Vec<CompletionItemWidgetWrap>)>)>,
 
     list: gtk::ListBox,
     css_provider: gtk::CssProvider,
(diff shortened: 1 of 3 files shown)


src/ui/popupmenu/lazy_loader.rs
```diff
@@ -19,7 +19,8 @@ struct State {
     /// Once we're loaded some (or all) data, this closure gets called if
     /// one exists. The first value in the tuple can is indication on the
     /// number of items needed before calling the closure.
-    once_loaded: Option<(Option<i32>, Box<Fn(&Vec<CompletionItemWidgetWrap>)>)>,
+    once_loaded:
+        Option<(Option<i32>, Box<dyn Fn(&Vec<CompletionItemWidgetWrap>)>)>,
```
vhakulinen: Was the compiler giving you warnings here? For me (rustc 1.36) doesn't complain anything.

smolck: Yes, but only on nightly. Stable doesn’t complain, but nightly says `warning: trait objects without an explicit ‘dyn’ are deprecated` and then suggests to put a ` dyn` there. You can see it in the Travis nightly (and I think beta) checks, [see this one for master](https://travis-ci.org/vhakulinen/gnvim/jobs/567500562#L742).

PR #71, opened 2019-07-13
Update versions of dependencies

smolck:
As I said in the commit msg, a few places needed to be updated to use `Some` when they previously didn't. In addition to that, the syntax for using `None`, say to create a `gtk::Layout`, has changed. We now have to specify what the `None` is, via something like this (note the explicit `None::<>` instead of `None`):
```rust
let layout = gtk::Layout(None::<&gtk::Adjustment>, None::<&gtk::Adjustment>)
```
I may be wrong in my understanding about this change (my understanding is from reading related issues and the following blog post), but I couldn't compile GNvim without these more explicit annotations, so I've put them in. Read more about this change (and others) here: https://gtk-rs.org/blog/2019/06/22/new-release.html

Let me know what you think (and if I missed a dependency, if we can't update for compatibility/similar reasons, etc.); any feedback is welcome. Note that I left the `features=` part of the Cargo.toml file for webkit2gtk and gtk as is, not sure if we want to change that (if so let me know).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,11 +7,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "ammonia"
-version = "2.0.0"
+version = "2.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "html5ever 0.22.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "lazy_static 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "html5ever 0.23.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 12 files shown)


badosu:
Tested locally, could not find any issues for my workload :clap:

smolck:
So in the Cargo.toml file [here](https://github.com/vhakulinen/gnvim/blob/master/Cargo.toml#L3), the version for GNvim isn't updated to what the current release is. Should I update it in this PR to be the current release version (0.1.4), the next release version (0.1.5, since it's the master branch), or leave it as is? I figure since I'm already updating the Cargo.toml I might as well fix that, since it's a small issue (and fix).

vhakulinen:
The version number in Cargo.toml isn't used currently anywhere. I'm not sure if it is required or not, but if not it should probably be removed. The version number to the gnvim binary is extracted from git (in build.rs).

PR #72, opened 2019-07-13
Fix `RedrawEvent::GridScroll`

### LINKED ISSUE #63: scrolloff setting cases line rendering problems
Hi there,

I'm getting rendering issues that I'm not sure how to exactly describe what I saw in words.

## How to reproduce

1. `gnvim -- -u NONE`
2. Open up document longer than your screen height.
3. `:set scrolloff=999`
4. Scroll to a location your cursor is in the middle of the screen area.
5. Hit `o` and hit Enter a few times.

## Other notes

1. This does not happen if one has `set number`. I reckon that the reason is cause the lines are redrawn to render the line numbers and so it shadows the problem.

#### gnvim version -- 0.1.3

smolck:
Fixes #63 (from my tests).

This PR comes from discussion on [GNvim's gitter](https://gitter.im/gnvim/community) about grid scrolling, where it was mentioned that when given a `grid_scroll` event like the following GNvim should execute each event in order: 
`["grid_scroll", [1, 0, 29, 0, 116, 5, 0], [1, 10, 29, 0, 116, -5, 0]]`

Previously, GNvim would only execute the first event, which caused #63. Now that should be fixed.

Let me know if this is how we want to solve this problem. This solution loops through each `grid_scroll` event, appends the info for each one to a vector (via the tuple struct `GridScrollInfo`), and delivers that vector to the UI to handle each event, but it isn't perfect. Any feedback on alternatives (or anything else) is definitely welcome.

### DIFF
--- a/src/nvim_bridge/mod.rs
+++ b/src/nvim_bridge/mod.rs
@@ -327,6 +327,10 @@ pub struct CmdlineShow {
 }
 
 #[derive(Debug, PartialEq)]
+/// grid, [top, bot, left, right], rows, cols
+pub struct GridScrollInfo(pub u64, pub [u64; 4], pub i64, pub i64);
+
+#[derive(Debug, PartialEq)]
 pub enum RedrawEvent {
     SetTitle(String),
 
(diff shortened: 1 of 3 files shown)


vhakulinen:
While this fixes #63, I think we should fix the underlying issue: all the enum members of `RedrawEvent` should have `Vec` as data. I'll try to find some time today/this week and see if there are any nice ways to adjust the current solution to achieve that, or are we facing a rewrite on the redraw event parsing/handling.

smolck:
@vhakulinen Okay, cool! I am happy to help out here, but should we use this PR for the refactor, or another one? (Probably not this one if we are refactoring the whole redraw event parsing/handling?) I can always close this one in favor of another if need be.

vhakulinen:
Merged. I'll try to open an issue about the underlying problem and try to list my ideas there. That can also serve as initial discussion place for the refactoring. Meanwhile, feel free to hack on it.

PR #86, opened 2019-07-31
Fix inefficiency with ui attachment

### LINKED ISSUE #82: Initial grid size is too small
I use the `sway` Wayland compositor. When I initially open `gnvim`, the grid size is smaller than the actual window size, and I need to resize it for it to fix itself.

Right after opening:
image

After resizing to fix the issue:
image

This also happens with smaller window sizes:
image

smolck:
Bfredl mentioned that there was an inefficency in how GNvim initializes the UI in https://github.com/vhakulinen/gnvim/issues/82#issuecomment-514729568. This commit fixes that according to his suggestion [here](https://github.com/vhakulinen/gnvim/issues/82#issuecomment-516729048).

This fixes #82 for me (with pango 1.43 at least; since pango 1.44 causes problems with GNvim, I downgraded pango to test if it fixed the issue). Apparently that slight inefficiency was the problem (Thanks @bfredl! 👏). That does need confirmation though, so if anyone can test this please do.

This isn't perfect, so feedback is appreciated; I imagine I probably missed something with the references and such, so if there's anything wrong there (or anywhere else), please let me know.

The current problem with this ATM is that none of the command line arguments to disable external features work, since I have defaulted to setting the external parts of the UI to true. Let me know how we want to go about fixing that.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,7 +20,7 @@ extern crate webkit2gtk;
 
 use gio::prelude::*;
 
-use neovim_lib::neovim::{Neovim, UiAttachOptions};
+use neovim_lib::neovim::Neovim;
 use neovim_lib::session::Session as NeovimSession;
 use neovim_lib::NeovimApi;
 
@@ -120,17 +120,6 @@ fn build(app: &gtk::Application, opts: &Options) {
     let api_info = nvim.get_api_info().expect("Failed to get API info");
     nvim.set_var("gnvim_channel_id", api_info[0].clone())
         .expect("Failed to set g:gnvim_channel_id");
-
-    let mut ui_opts = UiAttachOptions::new();
-    ui_opts.set_rgb(true);
-    ui_opts.set_linegrid_external(true);
-    ui_opts.set_popupmenu_external(!opts.disable_ext_popupmenu);
-    ui_opts.set_tabline_external(!opts.disable_ext_tabline);
-    ui_opts.set_cmdline_external(!opts.disable_ext_cmdline);
-
-    ui_opts.set_wildmenu_external(true);
-    nvim.ui_attach(80, 30, &ui_opts)
-        .expect("Failed to attach UI");
 
     let ui = ui::UI::init(app, rx, Arc::new(Mutex::new(nvim)));
     ui.start();
--- a/src/ui/ui.rs
+++ b/src/ui/ui.rs
@@ -7,7 +7,7 @@ use std::time;
 use gdk;
 use glib;
 use gtk;
-use neovim_lib::neovim::Neovim;
+use neovim_lib::neovim::{Neovim, UiAttachOptions};
 use neovim_lib::neovim_api::NeovimApi;
 use neovim_lib::NeovimApiAsync;
 use neovim_lib::Value;
@@ -74,6 +74,8 @@ struct UIState {
 
     /// Source id for delayed call to ui_try_resize.
     resize_source_id: Arc<Mutex<Option<glib::SourceId>>>,
+
+    pub attached: bool,
 }
 
 /// Main UI structure.
@@ -129,52 +131,6 @@ impl UI {
         // Create default grid.
         let mut grid = Grid::new(1);
         box_.pack_start(&grid.widget(), true, true, 0);
-
-        // When resizing our window (main grid), we'll have to tell neovim to
-        // resize it self also. The notify to nvim is send with a small delay,
-        // so we don't spam it multiple times a second. source_id is used to
-        // track the function timeout.
-        let source_id = Arc::new(Mutex::new(None));
-        let source_id_ref = source_id.clone();
-        let nvim_ref = nvim.clone();
-        grid.connect_da_resize(move |rows, cols| {
-            let nvim_ref = nvim_ref.clone();
-
-            let source_id_moved = source_id_ref.clone();
-            // Set timeout to notify nvim about the new size.
-            let new = glib::timeout_add(30, move || {
-                let mut nvim = nvim_ref.lock().unwrap();
-                nvim.ui_try_resize_async(cols as i64, rows as i64)
-                    .cb(|res| {
-                        if let Err(err) = res {
-                            eprintln!("Error: failed to resize nvim when grid size changed ({:?})", err);
-                        }
-                    })
-                .call();
-
-                // Set the source_id to none, so we don't accidentally remove
-                // it since it used at this point.
-                let source_id = source_id_moved.clone();
-                let mut source_id = source_id.lock().unwrap();
-                *source_id = None;
-
-                Continue(false)
-            });
-
-            let source_id = source_id_ref.clone();
-            let mut source_id = source_id.lock().unwrap();
-            {
-                // If we have earlier timeout, remove it.
-                let old = source_id.take();
-                if let Some(old) = old {
-                    glib::source::source_remove(old);
-                }
-            }
-
-            *source_id = Some(new);
-
-            false
-        });
 
         // Mouse button press event.
         let nvim_ref = nvim.clone();
@@ -280,7 +236,8 @@ impl UI {
         let mut grids = HashMap::new();
         grids.insert(1, grid);
 
-        UI {
+        let source_id = Arc::new(Mutex::new(None));
+        let ui = UI {
             win: Arc::new(ThreadGuard::new(window)),
             rx,
             state: Arc::new(ThreadGuard::new(UIState {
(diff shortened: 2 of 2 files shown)


smolck:
I noticed that Travis was failing on stable Rust, and after some investigating I think I know why. TL;DR is that Travis stable is using an outdated version of Rust. 

Reason the build fails is, I think, due to the borrow checker of Rust v1.35.0 (which Travis uses) not being able to understand how this branch borrows the ui state and then drops the reference to it before the end of the function so that the UI can be returned. In the new update to Rust (v1.36.0), NLL (Non-Lexical Lifetimes) was added (see [this blog post on the update](https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html#nll-for-rust-2015)), making the borrow checker smarter; I think this is why the beta and nightly versions of Rust compile and pass the Travis checks without any issues, but the stable channel doesn't.

Since this branch requires the latest version of Rust, should we find a workaround/alternative here that doesn't use `drop`? I'm not certain using `drop` is the best way to do this regardless, so I'm open to any alternatives.

src/main.rs
```diff
@@ -121,17 +121,6 @@ fn build(app: &gtk::Application, opts: &Options) {
     nvim.set_var("gnvim_channel_id", api_info[0].clone())
         .expect("Failed to set g:gnvim_channel_id");
 
-    let mut ui_opts = UiAttachOptions::new();
```
badosu:
```
warning: unused import: `UiAttachOptions`
  --> src/main.rs:23:34
   |
23 | use neovim_lib::neovim::{Neovim, UiAttachOptions};
   |                                  ^^^^^^^^^^^^^^^
   |
   = note: #[warn(unused_imports)] on by default
```

badosu:
Can't reproduce on this branch :+1:

src/main.rs
smolck: Oops, thanks! Fixed.

vhakulinen:
This doesn't currently work with gnome. This is because you're missing the first draw signal which happens in `window.show_all`.

smolck:
@vhakulinen Hmm. I tried this, but it didn't work right (as mentioned below):
```diff
diff --git a/src/ui/ui.rs b/src/ui/ui.rs
index f416ff8..3d89e1b 100644
--- a/src/ui/ui.rs
+++ b/src/ui/ui.rs
@@ -226,13 +226,8 @@ impl UI {
         let cmdline = Cmdline::new(&overlay, nvim.clone());
         let cursor_tooltip = CursorTooltip::new(&overlay);
 
-        window.show_all();
-
         grid.set_im_context(&im_context);
 
-        cmdline.hide();
-        cursor_tooltip.hide();
-
         let mut grids = HashMap::new();
         grids.insert(1, grid);
 
@@ -323,6 +318,10 @@ impl UI {
             false
         });
 
+        ui.win.borrow().show_all();
+        ui.state.borrow().cmdline.hide();
+        ui.state.borrow().cursor_tooltip.hide();
+
         ui
     }
```

It seems that the `window.show_all()` call is a bit finicky, and varies depending on position (I'm sure there's a reason for this, but I don't know it; I imagine you might, so if you can elaborate on this please do). This is what GNvim looks like when using the above diff (run using `GNVIM_RUNTIME_PATH=... cargo run`): 
image
Not sure why there's that white part, or why that messes up the grid's initial size; any ideas?

vhakulinen:
I managed to get a fix for this. I'll push it once I'm happy with the code.

smolck:
@vhakulinen @badosu So concerning the inefficiency that still exists with the UI attachment, how should we go about fixing that issue? More specifically, is this branch good groundwork or should we think of a new method of handling the issue?

vhakulinen:
I'm not sure if that needs fixing at the moment. Its wasted CPU cycles for sure, but for now I would concentrate efforts on fixing bugs.

That said, if you _want_ to fix this, there is a way: after the `window.show_all` call, attach a `gdk::idle_add` callback where you calculate the initial screen size and call `uiattach` on nvim. After the attachment, register the current resize handler.

bfredl:
> I'm not sure if that needs fixing at the moment. Its wasted CPU cycles for sure, but for now I would concentrate efforts on fixing bugs.

This certainly causes more issues than wasting CPU cycles, which never was the main issue.  If there is an error during `init.vim`, chances are the error prompt will be shown relative the first incorrect screen size, and not the true one, and error messages might get truncated/hidden etc. Not wasting CPU cycles is a result of doing things right, not the other way around.

"Ideally" nvim should support `ui_try_resize` at any time, and have messages semantically redrawn to the new screen size, but this work is not done yet (essentially it would mean using `ext_messages` internally in nvim, rather than storing messages directly on the grid).

smolck:
> That said, if you _want_ to fix this, there is a way: after the `window.show_all` call, attach a `gdk::idle_add` callback where you calculate the initial screen size and call `uiattach` on nvim. After the attachment, register the current resize handler.

@vhakulinen Okay, thanks! I think I’ll go ahead and fix it (with that info it shouldn’t take too long). Do you want me to push the changes to this branch?

vhakulinen:
Open another PR.

smolck:
@vhakulinen Is this like what you meant?
```diff
diff --git a/src/main.rs b/src/main.rs
index 7928bf2..fd8dec3 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -129,10 +129,10 @@ fn build(app: &gtk::Application, opts: &Options) {
     ui_opts.set_cmdline_external(!opts.disable_ext_cmdline);
 
     ui_opts.set_wildmenu_external(true);
-    nvim.ui_attach(80, 30, &ui_opts)
-        .expect("Failed to attach UI");
+    // nvim.ui_attach(80, 30, &ui_opts)
+    //     .expect("Failed to attach UI");
 
-    let ui = ui::UI::init(app, rx, Arc::new(Mutex::new(nvim)));
+    let ui = ui::UI::init(app, rx, Arc::new(Mutex::new(nvim)), ui_opts);
     ui.start();
 }
 
diff --git a/src/ui/ui.rs b/src/ui/ui.rs
index d56a852..ae83112 100644
--- a/src/ui/ui.rs
+++ b/src/ui/ui.rs
@@ -11,6 +11,7 @@ use neovim_lib::neovim::Neovim;
 use neovim_lib::neovim_api::NeovimApi;
 use neovim_lib::NeovimApiAsync;
 use neovim_lib::Value;
+use neovim_lib::UiAttachOptions;
 
 use gtk::prelude::*;
 
@@ -109,6 +110,7 @@ impl UI {
         app: &gtk::Application,
         rx: Receiver<Message>,
         nvim: Arc<Mutex<Neovim>>,
+        nvim_attach_opts: UiAttachOptions,
     ) -> Self {
         // Create the main window.
         let window = gtk::ApplicationWindow::new(app);
@@ -139,50 +141,6 @@ impl UI {
         let mut grid = Grid::new(1);
         box_.pack_start(&grid.widget(), true, true, 0);
 
-        // When resizing our window (main grid), we'll have to tell neovim to
-        // resize it self also. The notify to nvim is send with a small delay,
-        // so we don't spam it multiple times a second. source_id is used to
-        // track the function timeout. This timeout might be canceled in
-        // redraw even handler if we receive a message that changes the size
-        // of the main grid.
-        let source_id = Arc::new(Mutex::new(None));
-        let source_id_ref = source_id.clone();
-        let nvim_ref = nvim.clone();
-        grid.connect_da_resize(move |rows, cols| {
-            let nvim_ref = nvim_ref.clone();
-
-            let source_id_moved = source_id_ref.clone();
-            // Set timeout to notify nvim about the new size.
-            let new = glib::timeout_add(30, move || {
-                let mut nvim = nvim_ref.lock().unwrap();
-                nvim.ui_try_resize_async(cols as i64, rows as i64)
-                    .cb(|res| {
-                        if let Err(err) = res {
-                            eprintln!("Error: failed to resize nvim when grid size changed ({:?})", err);
-                        }
-                    })
-                .call();
-
-                // Set the source_id to none, so we don't accidentally remove
-                // it since it used at this point.
-                let source_id = source_id_moved.clone();
-                let mut source_id = source_id.lock().unwrap();
-                *source_id = None;
-
-                Continue(false)
-            });
-
-            let source_id = source_id_ref.clone();
-            let mut source_id = source_id.lock().unwrap();
-            // If we have earlier timeout, remove it.
-            if let Some(old) = source_id.take() {
-                glib::source::source_remove(old);
-            }
-
-            *source_id = Some(new);
-
-            false
-        });
 
         // Mouse button press event.
         let nvim_ref = nvim.clone();
@@ -280,6 +238,66 @@ impl UI {
 
         window.show_all();
 
+        let (cols, rows) = grid.calc_size();
+        let nvim_ref = nvim.clone();
+        gtk::idle_add(move || {
+            nvim_ref
+                .lock()
+                .unwrap()
+                .ui_attach(rows,
+                           cols,
+                           &nvim_attach_opts)
+                .unwrap();
+
+            glib::Continue(false)
+        });
+
+        // When resizing our window (main grid), we'll have to tell neovim to
+        // resize it self also. The notify to nvim is send with a small delay,
+        // so we don't spam it multiple times a second. source_id is used to
+        // track the function timeout. This timeout might be canceled in
+        // redraw even handler if we receive a message that changes the size
+        // of the main grid.
+        let source_id = Arc::new(Mutex::new(None));
+        let source_id_ref = source_id.clone();
+        let nvim_ref = nvim.clone();
+        grid.connect_da_resize(move |rows, cols| {
+            let nvim_ref = nvim_ref.clone();
+
+            let source_id_moved = source_id_ref.clone();
+            // Set timeout to notify nvim about the new size.
+            let new = glib::timeout_add(30, move || {
+                let mut nvim = nvim_ref.lock().unwrap();
+                nvim.ui_try_resize_async(cols as i64, rows as i64)
+                    .cb(|res| {
+                        if let Err(err) = res {
+                            eprintln!("Error: failed to resize nvim when grid size changed ({:?})", err);
+                        }
+                    })
+                .call();
+
+                // Set the source_id to none, so we don't accidentally remove
+                // it since it used at this point.
+                let source_id = source_id_moved.clone();
+                let mut source_id = source_id.lock().unwrap();
+                *source_id = None;
+
+                Continue(false)
+            });
+
+            let source_id = source_id_ref.clone();
+            let mut source_id = source_id.lock().unwrap();
+            // If we have earlier timeout, remove it.
+            if let Some(old) = source_id.take() {
+                glib::source::source_remove(old);
+            }
+
+            *source_id = Some(new);
+
+            false
+        });
+
+
         grid.set_im_context(&im_context);
 
         cmdline.hide();
```

Only problem I've noticed is that the issue with Startify starting scrolled down (see https://github.com/vhakulinen/gnvim/issues/82#issuecomment-517123783) happens with the above diff.

vhakulinen:
You need to do the `calc_size` _inside_ the `idle_add` handler. You also need to attach the resize handler in the `idle_add` handler too after the nvim attach call. In the latest master commit (https://github.com/vhakulinen/gnvim/commit/6e62b64699994be30522926c0c861ac541d49e08), I introduced a macro to make the moving of clonable objects into closures a bit easier (basically copied a macro that gtk-rs examples has).

smolck:
@vhakulinen `Grid` doesn't implement `Clone`, which I think I need to use (right?). Should I put a `#[derive(Clone)]` above the `Grid` struct in grid.rs so that I can clone the grid, or is that method not ideal (it worked when I tested it, aside from the Startify problem).

vhakulinen:
You could use the `State` like you do in this current PR.

PR #95, opened 2019-08-15
Multigrid support

vhakulinen:


vhakulinen:
Yup, still WIP. Good to know that the floats crash, I don't really have any test cases for them (floats) on my own workflow currently.

vhakulinen:
Rebased.

badosu:
Let me know if I can QA once you deem ready for :+1:

vhakulinen:
Things should be now ready for testing. Some minor bugs still, like the messages window showing things from earlier renders if the grid size is not even and I noticed that `fzf` window has blinking cursor even tho' it also has a solid one (probably something to do with the terminal mode). Also, after a window split there might be some black lines/areas because of a bug in nvim.

badosu:
Awesome! Thanks for the heads up, will test along this week.

badosu:
The floating window position overlaps with the autocompletion using coc.vim (both when hint is above and below the cursor):

Screenshot from 2019-12-16 12-40-06

badosu:
I'm not sure yet how to reproduce it, but after a while switching to a tab reveals empty contents (it's not redrawn):

Screenshot from 2019-12-16 12-53-37

vhakulinen:
coc won't be able to figure out the position for the floating window because of the externalized popupmenu. You'll either have to somehow add logic to gnvim to expose the popupmenu position to nvim (and thus' to coc) or disable the externalized popupmenu.

smolck:
> You'll either have to add logic to gnvim to expose the popupmenu position to nvim . . .

@badosu I'll ask about this on the Neovim Gitter, and if I come up with something I'll PR it and let you know.

badosu:
@vhakulinen @smolck Thanks for checking, in neovim the positioning works fine.

I am not sure if this is an issue in neovim for this particular case (having externalized popupmenu and the completioninfo not interacting well), or something that neovim expects the GUI to deal with outside neovim's scope.

> I'm not sure yet how to reproduce it, but after switching to a tab, only half the buffer is redrawn.
>
> This is my current blocker on continuing with tests.

By the way, the issue above is a separate one and happens after less than 10 minutes of usage.

smolck:
> I am not sure if this is an issue in neovim for this particular case (having externalized popupmenu and the completioninfo not interacting well), or something that neovim expects the GUI to deal with outside neovim's scope.

@badosu After asking @bfredl on the Neovim Gitter about this, it seems it would take a (apparently not all that difficult) PR to add `relative='popupmenu'` to the floating window API, which I think would fix this issue.

vhakulinen:
Note to self: somewhere is a integer underflow. It happened while the msg window was constantly resizing (was doing `%s/something/something`).

vhakulinen:
Also, in FZF window the cursor is duplicated because the terminal buffer has its own solid cursor and gnvim is drawing its own cursor (and thats one column behind).

badosu:
I just tested again and the behavior where after dealing with splits, opening tabs, etc creates a bug where part of the buffer gets hidden is still present.

I'm not sure if this is related to FZF but I use it extensively to handle file opening.

Screenshot from 2020-01-13 14-26-46

bfredl:
I just tested this very briefly. most things regarding floats and external windows seems to be working. But when I use `call nvim_win_set_config(0, {'external': v:false, 'width':50, 'height':10})` i e convert an existing split window to an external window, the size of the window is wrong.

vhakulinen:
Good to know :+1:

vhakulinen:
Ugh. This code is quite messy and I don't remember enough of the details to clean it up, at least in timely manner... Anyways, this seems to be somewhat working now. Any tests cases for floating/external windows are more than welcome, if someone happens to have any.

vhakulinen:
This is now ready for testing. Going to merge soon.

vhakulinen:
Waiting for https://github.com/neovim/neovim/pull/12612.

badosu:
Awesome, I've been using a thin wrapper instead of gnvim lately after I unfortunately got the blur bug because of an update. Happy to test as soon as the PR is merged

vhakulinen:
> I've been using a thin wrapper instead of gnvim after I unfortunately got the blur bug because of an update.

The blurry scrolling thing related to pango 1.44 is fixed in master.

vhakulinen:
Rebased.

Issue #99, opened 2019-08-22
Improve CLI usage errors to hint about nvim vs. gnvim flags


dbarnett:
vim and neovim support a few flags for opening multiple files, `-d` for opening in side-by-side diff mode, `-o`/`-O` for opening multiple windows stacked vertically/horizontally, `-p` to open in multiple tabs, etc. gnvim supports opening splits and diffs from within nvim after starting it, but doesn't support those multiple-file options. If I try to pass flags like that when launching gnvim, I get errors like `error: Found argument '-d' which wasn't expected, or isn't valid in this context`.

Is it intentional not to support those flags, or could they be added?

smolck:
GNvim does support those commands; you just have to pass them to nvim, not GNvim, using `$ gnvim -- -<nvim option>`. Try running those commands like this:
```
$ gnvim -- -d <file1> <file2>
$ gnvim -- -o <file1> <file2>
```

The double dash is used to pass args to nvim; run `gnvim --help` for more info.

dbarnett:
Thanks, that does what I wanted!

I see now the subtle hint about how args are handled in the usage line `gnvim [FLAGS] [OPTIONS] [FILES]... [-- <ARGS>...]`. FWIW I would have figured it out on my own if the error a little more self-explanatory, like
```
$ gnvim -d file1 file2 
error: Found argument '-d' which wasn't expected, or isn't valid in this context. If this is an argument for nvim, try moving it after a -- separator.

USAGE:
    gnvim [GNVIM_FLAGS] [OPTIONS] [FILES]... [-- <NVIM_ARGS>...]

For more information try --help
```

smolck:
> FWIW I would have figured it out on my own if the error a little more self-explanatory, like . . .

Thanks for the info, I'll get a PR for this when I get the chance (unless someone else does it first).

dbarnett:
Sure thing, thanks!

And BTW one reason I've been going out of my way to report minor issues is I'm interested in eventually getting comfortable with Rust and contributing, so I'm poking around for things that bug me that I could solve. Hope it's not a nuisance. So far I'm somewhat interested in #70 which I'm running into on my Pixelbook, and possibly scrollbars and multigrid (#25). I'd also love to get some fancy navigation features like GUI menus, a [minimap](https://atom.io/packages/minimap), or marking placed signs directly on the scrollbar, but I'm getting the impression that gnvim specifically strives to _not_ get fancy and might discourage adding those kinds of features.

[...]

PR #101, opened 2019-08-27
Improve CLI Usage Error

smolck:
See #99.

Only current problem is that the error message says `error:` twice (on the first line):
```bash
error: error: Found argument '--test' which wasn't expected, or isn't valid in this context

USAGE:
    gnvim [FLAGS] [OPTIONS] [FILES]... [-- <ARGS>...]

For more information try --help
If this is an argument for nvim, try moving it after a -- separator.
```

I could do some string splitting to remove the double `error:` strings, but I doubt that's the best way. AFAIK it's caused by me using the description of the error [here](https://github.com/smolck/gnvim/blob/4b680aadbc6cb52a88c6f30c09a4f58c6c1ee5a6/src/main.rs#L149), and then calling `exit()` (which I think puts the `error:` part at the start of the description automatically). I could of course just put the usual message without the `error:` at the front into an `&str` (which is probably the best course of action; let me know what you think).

Any suggestions/feedback/etc. (especially on how to remove the extra `error:`) are appreciated.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -28,7 +28,7 @@ use std::cell::RefCell;
 use std::process::Command;
 use std::rc::Rc;
 
-use structopt::StructOpt;
+use structopt::{clap, StructOpt};
 
 include!(concat!(env!("OUT_DIR"), "/gnvim_version.rs"));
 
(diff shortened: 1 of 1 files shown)


vhakulinen:
Merged in https://github.com/vhakulinen/gnvim/commit/5687ffce0b1f24b9e4c8aa1ba6336c5bd10426a4

PR #102, opened 2019-08-29
Fix panic when cell_at fails sometimes

llwwns:
`cell_at` caused panic sometimes when I was typing Japanese words.

I don't know what is the real reason but it can be fixed by remove the `unwrap` and turn the return value of `cell_at` into optional.

Issue #92, opened 2019-08-13
Panic when using coc.nvim :CocList command


juho-p:
nvim version is 0.3.8. gnvim is compiled from current master.

Panic happens using following steps

- install coc.nvim plugin: https://github.com/neoclide/coc.nvim
- run :CocList -command

Terminal nvim works fine. I'm not sure exactly what coc.nvim is doing there.

Gnvim outputs following (including stacktrace). After typing `RUST_BACKTRACE=1 gnvim` command I immediately typed `CocList<cr>`. Crashes every time.

```
juho@juho-ThinkPad-P1:~$ RUST_BACKTRACE=1 gnvim
Not supported option set: arabicshape
Not supported option set: ambiwidth
Not supported option set: emoji
Not supported option set: guifontset
Not supported option set: guifontwide
Not supported option set: showtabline
Not supported option set: termguicolors
Not supported option set: ext_linegrid
Not supported option set: ext_hlstate
Not supported option set: ext_cmdline
Not supported option set: ext_popupmenu
Not supported option set: ext_tabline
Not supported option set: ext_wildmenu
Not supported option set: termguicolors
Not supported option set: ext_cmdline
Not supported option set: ext_popupmenu
Not supported option set: ext_tabline
Not supported option set: ext_wildmenu
Not supported option set: ext_cmdline
Not supported option set: ext_popupmenu
Not supported option set: ext_tabline
Not supported option set: ext_wildmenu
Not supported option set: ext_cmdline
Not supported option set: ext_popupmenu
Not supported option set: ext_tabline
Not supported option set: ext_wildmenu
Not supported option set: ext_cmdline
Not supported option set: ext_popupmenu
Not supported option set: ext_tabline
Not supported option set: ext_wildmenu
Failed to turn input event into nvim key (keyval: 65506)
Failed to turn input event into nvim key (keyval: 65506)
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:347:21
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:59
             at src/libstd/panicking.rs:197
   3: std::panicking::default_hook
[...]

PR #103, opened 2019-08-31
Fix panic at double-width leaf cell_at called

llwwns:
https://github.com/vhakulinen/gnvim/pull/102#issuecomment-526799751

### DIFF
--- a/src/ui/grid/row.rs
+++ b/src/ui/grid/row.rs
@@ -595,7 +595,13 @@ mod tests {
     fn test_rope_cell_at() {
         let left = Rope::Leaf(Leaf::new(String::from("123"), 0, false));
         let right = Rope::Leaf(Leaf::new(String::from("456"), 1, false));
-        let rope = Rope::Node(Box::new(left), Box::new(right));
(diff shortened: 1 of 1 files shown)


vhakulinen:
Add a test case covering this. Also, can you provide a (real) case where this breaks?

llwwns:
I find a way to make the panic happen.

1. Use this init.vim
```
set runtimepath+=~/.vim
call plug#begin('~/.vim/plugged')
Plug 'tyru/eskk.vim'
call plug#end()
```
eskk.vim is a Japanese IME written in vimscript.

2. Run
```
curl -fLo ~/.vim/autoload/plug.vim --create-dirs \
    https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
```

3. Run neovim and execute `Plug Install`

4. Rerun neovim and type <c-j>A<c-j>

smolck:
@vhakulinen Now that this is merged, I believe #117 and #92 should now be fixed (they panicked at the `cell_at` call),

vhakulinen:
Added comment there to wait for confirmation.

PR #106, opened 2019-09-15
Fix #41

### LINKED ISSUE #41: Feature request: option to show `menu` on inactive popupmenu completion items
Gnvim's externalized popupmenu currently only show's `info` and `menu` items on the active completion item. There should be an option to show `menu` on inactive once too.

smolck:
This uses a flag in the `Popupmenu` struct which can be set via RPC to determine whether or not the `menu` part of each _inactive_ (and active as well in most cases) `CompletionItem` is visible or not:

After `:call gnvim#popupmenu#show_menu_on_all_items("true")` has been run, no selection:

image

Same as above but when __selecting an item and pressing `<C-s>`__:

image

Same as above, but when __selecting an item without `<C-s>` being pressed__:

image

Currently the only issue I've noticed with this branch has been that when scrolling downwards _with the mouse_, things aren't updated until selected (which I'll look into hopefully later today, I have an idea of how to fix it):

image

Any feedback (especially on the variable/function names I've used) is greatly appreciated. Not sure if a flag in the `Popupmenu` struct is how we want to do this, so if it's not let me know.

See #41.

### DIFF
--- a/src/ui/popupmenu/popupmenu.rs
+++ b/src/ui/popupmenu/popupmenu.rs
@@ -77,6 +77,8 @@ pub struct Popupmenu {
     /// Label for displaying full info of a completion item.
     info_label: gtk::Label,
 
+    menu_on_all_items: bool,
+
     state: Rc<RefCell<State>>,
     items: LazyLoader,
 
@@ -256,6 +258,7 @@ impl Popupmenu {
 
         Popupmenu {
             items: LazyLoader::new(list.clone(), css_provider.clone()),
+            menu_on_all_items: false,
             box_,
             layout,
             css_provider,
@@ -269,6 +272,18 @@ impl Popupmenu {
             font: Font::default(),
             line_space: 0,
         }
+    }
+
+    pub fn set_menu_on_all_items(&mut self, b: bool) {
+        self.menu_on_all_items = b;
+    }
+
+    pub fn show_menu_for_all_items(&mut self) {
+        self.items.once_loaded(Some(5), move |items| {
+            for item in items {
+                item.menu.set_visible(true);
+            }
+        });
     }
 
     pub fn is_above_anchor(&self) -> bool {
@@ -348,7 +363,11 @@ impl Popupmenu {
     }
 
     /// Shows the popupmenu.
-    pub fn show(&self) {
+    pub fn show(&mut self) {
+        if self.menu_on_all_items {
+            self.show_menu_for_all_items();
+        }
+
         self.layout.show();
         self.box_.check_resize();
     }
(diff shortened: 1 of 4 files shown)


src/ui/ui.rs
```diff
@@ -671,11 +674,12 @@ fn handle_redraw_event(
                         grid.get_rect_for_cell(popupmenu.row, popupmenu.col);
 
                     state.popupmenu.set_anchor(rect);
-                    state.popupmenu.show();
                     state
                         .popupmenu
                         .select(popupmenu.selected as i32, &state.hl_defs);
 
+                    state.popupmenu.show();
```
smolck: Moved this because of the info in [this comment](https://github.com/vhakulinen/gnvim/blob/68e3fc8d2e8f6f7bbdd73641b00031c16004e8cc/src/ui/popupmenu/lazy_loader.rs#L134-L135) (since there may be a callback attached by `popupmenu.show()` when it is called now; see [`show_menu_for_all_items`](https://github.com/smolck/gnvim/blob/94cb9d476fa5bd198bbf032ecf110e029ff65d15/src/ui/popupmenu/popupmenu.rs#L281) which may be called [here](https://github.com/smolck/gnvim/blob/94cb9d476fa5bd198bbf032ecf110e029ff65d15/src/ui/popupmenu/popupmenu.rs#L368) in the `popupmenu.show` function):

> Only one callback can exists at a time (e.g. when we are loading items).

Is what I've done here okay, or is there a better way? What I've done works, but it feels a bit off to me.

src/ui/popupmenu/popupmenu.rs
```diff
@@ -271,6 +274,18 @@ impl Popupmenu {
         }
     }
 
+    pub fn set_menu_on_all_items(&mut self, b: bool) {
+        self.menu_on_all_items = b;
+    }
+
+    pub fn show_menu_for_all_items(&mut self) {
```
smolck: Again, any better names here?

src/ui/popupmenu/popupmenu.rs
```diff
@@ -271,6 +274,18 @@ impl Popupmenu {
         }
     }
 
+    pub fn set_menu_on_all_items(&mut self, b: bool) {
```
smolck: Needs a better name; the intent of the function isn't really obvious.

src/ui/popupmenu/popupmenu.rs
```diff
@@ -256,6 +258,7 @@ impl Popupmenu {
 
         Popupmenu {
             items: LazyLoader::new(list.clone(), css_provider.clone()),
+            menu_on_all_items: false,
```
smolck: Will set this to default to `true`, as per @vhakulinen in https://github.com/vhakulinen/gnvim/issues/41#issuecomment-530921001.

vhakulinen:
Thanks. Made some changes and merged.

clason:
I think the documentation is wrong: vimscript "bools" are `1` and `0`, strings have no truthiness. So you need to 
`:call gnvim#popupmenu#show_menu_on_all_items(1)`; `"true"` doesn't work.

vhakulinen:
That is mentioned in the documentation: https://github.com/vhakulinen/gnvim/blob/75f6e175a438d94eafd0ca066baa1bed809c038e/runtime/doc/gnvim.txt#L106-L109

clason:
True, but I thought it'd be good to mention this here as well in case someone comes here first via Google. (The documentation wasn't included in this PR, so the only instructions I saw were in the PR description.)

PR #109, opened 2019-09-16
Add geometry command line option

### LINKED ISSUE #108: Ability to set default size
There currently doesn't seem to be a way to set the size of the gnvim window at startup; it always defaults to 1280x720. (Differently from neovim-qt, the window size also isn't saved and restored when closing and reopening.)

Or am I missing something?

clason:
Closes #108 

Window size can now be specified using the `--geometry` option; defaults to `1280x720` as before. I know zero Rust, so I'm sure this could be done more elegantly.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -36,11 +36,23 @@ mod nvim_bridge;
 mod thread_guard;
 mod ui;
 
+fn parse_geometry(input: &str) -> Result<(i32, i32), String> {
+    let ret_tuple: Vec<&str> = input.split("x").collect();
+    if ret_tuple.len() != 2 {
+        Err(String::from("must be of form 'width'x'height'"))
+    } else {
+        match (ret_tuple[0].parse(), ret_tuple[1].parse()) {
+            (Ok(x), Ok(y)) => Ok((x, y)),
+            (_, _) => Err(String::from("at least one argument wasn't an integer")),
+        }
+    }
+}
+
 /// Gnvim is a graphical UI for neovim.
 #[derive(StructOpt, Debug)]
 #[structopt(
     name = "gnvim",
-    raw(version = "VERSION"),
+    version = VERSION,
     author = "Ville Hakulinen"
 )]
 struct Options {
@@ -79,6 +91,10 @@ struct Options {
     /// Disables externalized tab line
     #[structopt(long = "disable-ext-tabline")]
     disable_ext_tabline: bool,
+
+    /// Geometry of the window in widthxheight form
+    #[structopt(long = "geometry", parse(try_from_str = parse_geometry), default_value = "1280x720")]
+    geometry: (i32, i32),
 }
 
 fn build(app: &gtk::Application, opts: &Options) {
@@ -132,7 +148,7 @@ fn build(app: &gtk::Application, opts: &Options) {
     nvim.ui_attach(80, 30, &ui_opts)
         .expect("Failed to attach UI");
 
-    let ui = ui::UI::init(app, rx, Rc::new(RefCell::new(nvim)));
+    let ui = ui::UI::init(app, rx, opts.geometry, Rc::new(RefCell::new(nvim)));
     ui.start();
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -43,7 +43,9 @@ fn parse_geometry(input: &str) -> Result<(i32, i32), String> {
     } else {
         match (ret_tuple[0].parse(), ret_tuple[1].parse()) {
             (Ok(x), Ok(y)) => Ok((x, y)),
-            (_, _) => Err(String::from("at least one argument wasn't an integer")),
+            (_, _) => {
+                Err(String::from("at least one argument wasn't an integer"))
+            }
         }
     }
 }
--- a/src/ui/ui.rs
+++ b/src/ui/ui.rs
@@ -105,12 +105,13 @@ impl UI {
     pub fn init(
         app: &gtk::Application,
         rx: glib::Receiver<Message>,
+        window_size: (i32, i32),
         nvim: Rc<RefCell<Neovim>>,
     ) -> Self {
         // Create the main window.
         let window = gtk::ApplicationWindow::new(app);
         window.set_title("Neovim");
-        window.set_default_size(1280, 720);
+        window.set_default_size(window_size.0, window_size.1);
 
         // Top level widget.
         let b = gtk::Box::new(gtk::Orientation::Vertical, 0);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -314,7 +314,7 @@ dependencies = [
  "pango 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "pangocairo 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "pulldown-cmark 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "structopt 0.2.18 (registry+https://github.com/rust-lang/crates.io-index)",
+ "structopt 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "syntect 3.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "webkit2gtk 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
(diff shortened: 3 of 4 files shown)


src/main.rs
```diff
@@ -132,7 +136,9 @@ fn build(app: &gtk::Application, opts: &Options) {
     nvim.ui_attach(80, 30, &ui_opts)
         .expect("Failed to attach UI");
 
-    let ui = ui::UI::init(app, rx, Rc::new(RefCell::new(nvim)));
+    let size = opts.geometry.split("x").collect();
```
smolck: Maybe call this `window_size` instead?

src/ui/ui.rs
```diff
@@ -105,12 +105,13 @@ impl UI {
     pub fn init(
         app: &gtk::Application,
         rx: glib::Receiver<Message>,
+        size: Vec<&str>,
```
smolck: Same as above. Also, maybe change it to be a tuple of `(i32, i32)` and then parse it on the other side (in `main.rs`).

src/ui/ui.rs
```diff
@@ -105,12 +105,13 @@ impl UI {
...
         nvim: Rc<RefCell<Neovim>>,
     ) -> Self {
         // Create the main window.
         let window = gtk::ApplicationWindow::new(app);
         window.set_title("Neovim");
-        window.set_default_size(1280, 720);
+        window.set_default_size(size[0].parse().unwrap(), size[1].parse().unwrap());
```
smolck: Probably should parse things with `size[0].parse().unwrap_or(1280)` and `size[1].parse().unwrap_or(720)` instead, so we don’t crash GNvim if the parse fails (assuming it can fail if the user passes a bad arg; I’m not 100% sure it can). We may want to notify the user too.

src/main.rs
```diff
@@ -68,6 +68,10 @@ struct Options {
     #[structopt(value_name = "ARGS", last = true)]
     nvim_args: Vec<String>,
 
+    /// Window size
```
badosu: These comments are used to present the description of the option when `gnvim --help` is run.

Assume the user can read this comment and provide a more descriptive message, e.g.:

`Geometry of the window in <width>x<height> format

src/ui/ui.rs
clason: Sorry, I have to give up here. The critical point is the parsing of the string, and I can't figure out how to do proper try--catch error handling in Rust. A copy&pastable solution would be appreciated.

src/main.rs
```diff
@@ -132,7 +136,11 @@ fn build(app: &gtk::Application, opts: &Options) {
     nvim.ui_attach(80, 30, &ui_opts)
         .expect("Failed to attach UI");
 
-    let ui = ui::UI::init(app, rx, Rc::new(RefCell::new(nvim)));
+    let geometry = opts.geometry.split("x").collect::<Vec<&str>>();
+    let window_size = (geometry[0].parse::<i32>().unwrap(),
+                       geometry[1].parse::<i32>().unwrap());
```
smolck: To handle the possible errors here with parsing to `i32`’s just do this instead:
```rust
    let window_size = (geometry[0].parse::<i32>().unwrap_or(1280),
	               geometry[1].parse::<i32>().unwrap_or(720));
```

`unwrap_or(1280)` is essentially saying, “Hey, try and get a value from this. If it’s an error and not a value, just give me 1280,” and likewise for `unwrap_or(720)`.

clason: That's fine, but what happens if I do `--geometry 1280` or `--geometry whatever`? I'm trying to figure out how to fail gracefully on line 139.

clason: Also, I'd rather error out with `Invalid geometry!" than silently revert to a default.

smolck: Oh, right; sorry for misunderstanding the problem! What happens right now if you pass `--geometry whatever`?

clason:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', src/libcore/result.rs:1084:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
fish: “gnvim --geometry whateve” terminated by signal SIGILL (Illegal instruction)
```

badosu:
Great work @clason :clap:

src/main.rs
smolck: How about something like this:

```rust
    let window_size = (geometry[0].parse::<i32>().unwrap_or_else(panic!(“Invalid geometry!”), ...)
```

or something like:

```rust
geometry[1].parse::<i32>().unwrap_or_else(|| {
        println!(“Invalid geometry! Defaulting to 1280x720”);
        720 // or maybe `return;` if we just want to exit GNvim
}));
```

clason: That's certainly nicer, but doesn't cover the `split` error (which is the hardest to catch).

smolck: Yeah, for sure. This is ugly, but I think it’d work:
```rust
if geometry.len() != 2 {
   println!(“Invalid geometry!”);
   return // Or whatever.
}
```

Or: 
```rust
 let window_size = if geometry.len() != 2 {
    (1280, 720)
} else {
    // Parse to i32’s and make into tuple and return it
}
```

badosu: `suggestion
    let window_size = (geometry[0].parse::<i32>().unwrap_or(1280),
	               geometry[1].parse::<i32>().unwrap_or(720));
`

clason: No, `unwrap_or_else` doesn't seem to work like this. (And it complains about `return` in a function that has return types.)

clason:
I now handle `split` failure with an assert and `parse` failure with a default. Both aren't ideal, but they work. Proper error handling has to wait for someone who actually knows Rust...

smolck:
This may be the best solution:
```diff
diff --git a/Cargo.toml b/Cargo.toml
index 578c2c9..8c92118 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ unstable = []
 
 [dependencies]
 #neovim-lib = "0.6"
-structopt = "0.2"
+structopt = "0.3.1"
 
 gio = "0.7"
 pango = "0.7"
diff --git a/src/main.rs b/src/main.rs
index 9b2db92..c6cf9cb 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -36,11 +36,22 @@ mod nvim_bridge;
 mod thread_guard;
 mod ui;
 
+fn parse_geometry(input: &str) -> Result<(i32, i32), String> {
+    let ret_tuple: Vec<&str> = input.split("x").collect();
+    if ret_tuple.len() != 2 {
+        Err(String::from("Invalid geometry!"))
+    } else {
+        match (ret_tuple[0].parse(), ret_tuple[1].parse()) {
+            (Ok(x), Ok(y)) => Ok((x, y)),
+            (_, _) => Err(format!("Invalid geometry, at least one arg wasn't a number.")),
+        }
+    }
+}
+
 /// Gnvim is a graphical UI for neovim.
 #[derive(StructOpt, Debug)]
 #[structopt(
     name = "gnvim",
-    raw(version = "VERSION"),
     author = "Ville Hakulinen"
 )]
 struct Options {
@@ -79,8 +90,12 @@ struct Options {
     /// Disables externalized tab line
     #[structopt(long = "disable-ext-tabline")]
     disable_ext_tabline: bool,
+
+    #[structopt(long = "geometry", parse(try_from_str = parse_geometry), default_value = "1280x720")]
+    geometry: (i32, i32),
 }
 
+
 fn build(app: &gtk::Application, opts: &Options) {
     let (tx, rx) = glib::MainContext::channel(glib::PRIORITY_DEFAULT);
 
```

Note that `raw` is deprecated except for when used with `raw(true)` or `raw(false)`, so that will have to be dealt with.

 When the above diff is used, it provides a nice error message (that probably needs some tweaking):
```
error: Invalid value for '--geometry <geometry>': Invalid geometry!
```

clason:
@smolck That's great, exactly what I was looking for! I've included your version. From what I've read, `raw` is deprecated because it's not needed anymore? 

(Bumping structop also pulled in more dependencies; I hope that's OK.)

smolck:
> From what I've read, raw is deprecated because it's not needed anymore?
(Bumping structop also pulled in more dependencies; I hope that's OK.)

The new dependencies should be fine. Not sure about `raw` though, haven’t looked into it.

One more thing; to clear up the commit history, would you mind doing an interactive rebase? e.g.
```
$ git rebase -i HEAD~5
...<‘squash’ everything into one commit>
$ git push -f
```

Not sure if you’re familiar with interactive rebasing. It isn’t strictly necessary, but it makes the commit history a bit nicer (one commit instead of 5). See https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History for more info.

Thanks again for the PR!

clason:
I assumed you'd be squash merging this anyway? That's usually preferred over rebasing. But sure, I can also rebase.

smolck:
> I assumed you'd be squash merging this anyway? That's usually preferred over rebasing. But sure, I can also rebase.

Huh, didn’t now that existed! @vhakulinen is the owner of this repo though, so I can’t merge this anyways. If you’d rather the PR be squash merged then he’d be the one to talk to.

clason:
It's an option in Github: https://help.github.com/en/articles/configuring-commit-squashing-for-pull-requests (can even be made default).

Rebased anyway (first try! \o/)

vhakulinen:
Merged.

Issue #110, opened 2019-09-16
Installing on OSX?


leeola:
I'm having trouble installing gnvim on OSX due to the libwebkit2gtk requirement. How crucial is that to operating gnvim, and could a feature be added to opt out of that requirement?

smolck:
@vhakulinen Isn’t libwebkit2gtk used to render markdown in the cursor tooltip?

badosu:
Does Rust support optional dependencies?

smolck:
> Does Rust support optional dependencies?
> 
> Edit: [Seems like it](https://github.com/rust-lang-nursery/edition-guide/issues/96). I think apart from the build process and feature handling adjustments, should be quite straightforward

Hopefully it can be done without too much boilerplate (e.g. `#cfg(...)` statements everywhere). Worst case scenario we could maybe make a separate, 'bare bones' branch of GNvim or something.

> BTW, I've never seen well formatted markdown on gnvim, when does it happen?

I believe it happens when hovering over something with gnvim-lsp installed; see bottom of [this page in the wiki](https://github.com/vhakulinen/gnvim/wiki) and [`cursor_tooltip.rs`](https://github.com/vhakulinen/gnvim/blob/master/src/ui/cursor_tooltip.rs).

smolck:
@leeola Maybe https://github.com/vhakulinen/gnvim/issues/8#issuecomment-517421794 can help you?

leeola:
I actually found that before posting _(during searching/etc)_, but didn't even try it due to it not working for the OP haha. Apparently it took over 2 hours, and still failed. I didn't bother heh.

leeola:
Appreciate the discussion all. I should note that this issue is not widely important to me _(or I'd offer to help)_, so don't try too much on my account. I just wanted to give Gnvim a try, and thought I was missing something stupid. Thanks :)

vhakulinen:
Adding the webkit2gtk dependency behind a (compile time) feature flag should be possible. PRs for this are welcome. Webkit is used to display markdown documents on the cursor tooltip.

Closing this as duplicate of #8.

Lythenas:
I'm working on a PR for this. I haven't worked with rust for a while and this seems easy to do.

kropper:
Yes, it builds now without error! :-)

However, after starting it, fonts are pretty messed up and the insert cursor cuts characters in half. Are there any settings recommended for running on macOS?

vhakulinen:
[...]

Issue #85, opened 2019-07-28
Font rendering issues with pango 1.44


last-partizan:
After updating to pango-1.44 command menu is invisible (i'm using `--disable-ext-cmdline`, with gnvim's cmdline it's okay)
![Screenshot from 2019-07-28 22-59-39](https://user-images.githubusercontent.com/301015/62012218-16594e00-b18c-11e9-9172-f9098395d762.png)

After scrolling with mouse scroll fonts become blurred:
![Screenshot from 2019-07-28 22-59-01](https://user-images.githubusercontent.com/301015/62012226-2bce7800-b18c-11e9-83bc-7ddd030c13de.png)

neovim-gtk also has similar problems https://github.com/daa84/neovim-gtk/issues/208 (and there is some details about why: pango deprecated Freetype and using Harfbuzz instead)

vhakulinen:
This is likely a duplicate of #84. I'll close this in favor of that.

vhakulinen:
Reopened this to track progress with the pango 1.44 issue. I miss interpreted #84 to be an issue with pango 1.44, which is not. There is some discussion in that (#84) that are related to this issue here.

smolck:
So, this is interesting. Supposedly, pango is now supposed to continue rounding glyph positions by default, maintaining the pre-1.44 behavior; subpixel positioning is now optional. However, when throwing a `dbg(&self)` after [this line in context.rs](https://github.com/vhakulinen/gnvim/blob/68e3fc8d2e8f6f7bbdd73641b00031c16004e8cc/src/ui/grid/context.rs#L216), I get the following:

Pango 1.43:
```
[src/ui/grid/context.rs:217] &self = CellMetrics {
    height: 26.0,
    width: 10.0,
    ascent: 19.0,
    decent: 7.0,
    underline_thickness: 2.0,
    underline_position: -1.0,
    line_space: 0,
    font: Font {
        name: "Iosevka Extrabold",
        height: 14.0,
    },
}
```

Pango ~1.44.5:
```
[src/ui/grid/context.rs:217] &self = CellMetrics {
    height: 24.5283203125,
    width: 9.0,
    ascent: 18.2373046875,
    decent: 6.291015625,
    underline_thickness: 1.8671875,
    underline_position: -0.93359375,
    line_space: 0,
    font: Font {
        name: "Iosevka Extrabold",
        height: 14.0,
    },
}
```

[...]

PR #112, opened 2019-09-17
Add feature flag for webkit

Lythenas:
Put `webkit2gtk` behind a feature flag (enabled by default) which removed the documentation tooltip.

Related  to #8 and #110.

This is the first attempt that compiled (and if it compiles it's correct, right?). It works for me locally but maybe someone else should try it. Also someone who uses a mac could try if they can compile gnvim using the `--no-default-features` flag.

I'm not familiar with gtk and this can probably be cleaned up a bit. Like not creating the tooltip at all if the feature is disabled instead of just not adding the webview.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,10 @@ authors = ["Ville Hakulinen <ville.hakulinen@gmail.com>"]
 debug = true
 
 [features]
+default = ["webkit"]
+
 unstable = []
+webkit = ["webkit2gtk"]
 
 [dependencies]
 #neovim-lib = "0.6"
@@ -33,6 +36,7 @@ features = ["v3_18"]
 [dependencies.webkit2gtk]
 version = "0.8"
 features = ["v2_14"]
+optional = true
 
 [dependencies.neovim-lib]
 #git = "https://github.com/vhakulinen/neovim-lib"
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,6 +16,7 @@ extern crate gtk;
 extern crate neovim_lib;
 extern crate pango;
 extern crate pangocairo;
+#[cfg(feature = "webkit")]
 extern crate webkit2gtk;
 
 use gio::prelude::*;
--- a/src/ui/cursor_tooltip.rs
+++ b/src/ui/cursor_tooltip.rs
@@ -7,7 +7,9 @@ use std::sync::Arc;
 use gtk;
 use gtk::prelude::*;
 
+#[cfg(feature = "webkit")]
 use webkit2gtk as webkit;
+#[cfg(feature = "webkit")]
 use webkit2gtk::{SettingsExt, WebViewExt};
 
 use ammonia;
@@ -82,6 +84,7 @@ pub struct CursorTooltip {
     css_provider: gtk::CssProvider,
     frame: gtk::Frame,
     fixed: gtk::Fixed,
+    #[cfg(feature = "webkit")]
     webview: webkit::WebView,
     state: Arc<ThreadGuard<State>>,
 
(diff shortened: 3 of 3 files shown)


smolck:
LGTM 👍

sebnow:
It compiles and runs for me. Hard to say how well it works though. I get graphical artifacts, and the cursor disappears sometimes. I don't know how it would work on Linux.
 
```
 $ PKG_CONFIG_PATH="/usr/local/opt/libffi/lib/pkgconfig" GNVIM_RUNTIME_PATH=$(pwd)/runtime cargo run --no-default-features
   Compiling gnvim v0.1.0 (/Users/sebnow/code/gnvim)
warning: unused variable: `frame`
   --> src/ui/cursor_tooltip.rs:124:27
    |
124 |             clone!(state, frame => move |fixed, alloc| {
    |                           ^^^^^ help: consider prefixing with an underscore: `_frame`
    |
    = note: #[warn(unused_variables)] on by default

warning: unused variable: `all`
   --> src/ui/cursor_tooltip.rs:366:13
    |
366 |         let all = format!(
    |             ^^^ help: consider prefixing with an underscore: `_all`

warning: method is never used: `borrow`
  --> src/thread_guard.rs:23:5
   |
23 |     pub fn borrow(&self) -> Ref<T> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: #[warn(dead_code)] on by default

warning: constant item is never used: `MAX_WIDTH`
  --> src/ui/cursor_tooltip.rs:50:1
   |
50 | const MAX_WIDTH: i32 = 700;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: constant item is never used: `MAX_HEIGHT`
  --> src/ui/cursor_tooltip.rs:51:1
   |
51 | const MAX_HEIGHT: i32 = 300;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    Finished dev [unoptimized + debuginfo] target(s) in 9.63s
     Running `target/debug/gnvim`
Not supported option set: arabicshape
Not supported option set: ambiwidth
Not supported option set: emoji
Not supported option set: guifontset
Not supported option set: guifontwide
Not supported option set: pumblend
Not supported option set: showtabline
Not supported option set: termguicolors
Not supported option set: ext_linegrid
Not supported option set: ext_multigrid
Not supported option set: ext_hlstate
Not supported option set: ext_termcolors
Received unknown redraw event: hl_group_set
Not supported option set: ext_cmdline
Not supported option set: ext_popupmenu
Not supported option set: ext_tabline
Not supported option set: ext_wildmenu
Not supported option set: ext_messages
Not supported option set: termguicolors
Received unknown redraw event: set_icon
Received unknown redraw event: hl_group_set
Received unknown redraw event: hl_group_set
Received unknown redraw event: set_icon
Not supported option set: ext_cmdline
Not supported option set: ext_popupmenu
Not supported option set: ext_tabline
Not supported option set: ext_wildmenu
Not supported option set: ext_messages
Failed to turn input event into nvim key (keyval: 65507)
Received unknown redraw event: set_icon
Received unknown redraw event: set_icon
Received unknown redraw event: set_icon
Received unknown redraw event: set_icon
Failed to turn input event into nvim key (keyval: 65506)
Failed to turn input event into nvim key (keyval: 65507)
Failed to turn input event into nvim key (keyval: 65506)
```

smolck:
@sebnow Could you show a screenshot of some of the graphical artifacts your getting? Is there a specific case when you get the artifacts or cursor disappearing (e.g. when the popupmenu shows, when you type in a file, etc.)?

sebnow:
Artifacts related to scrolling down:
Screen Shot 2019-09-18 at 15 35 17
Ignore the font rendering issues - I haven't played around with the options yet.

I was able to replicate the cursor disappearing by using [`:CocList files`](https://github.com/neoclide/coc-lists), going into insert mode, and escaping.

smolck:
What version of `pango` are you using (unless MacOS doesn’t have that; I’m assuming it does)? If you’re using version 1.44+ of `pango`, try downgrading it to 1.43 if possible. See #85. Not sure if that is the problem, but it looks like it to me.

badosu:
> > Ignore the font rendering issues - I haven't played around with the options yet.
>
> Don’t think any options can change those font issues (correct me if I’m wrong); they shouldn’t be happening, so I didn’t ignore them in my above response.

I've had similar issues on another machine I have Gnvim installed (Manjaro Arch derivate), I'll try to reproduce and verify this might not be a OSX specific issue.

Lythenas:
As mentioned above this is the first thing I got to compile. It is still displaying the cursor tooltip just without the webview. And I'm guessing then it should have size 0 but that might still lead to artifacts.

Because the cursor tooltip is only needed if webkit is available I will try to not display it at all.

If I understand correctly without this PR gnvim wouldn't compile on MacOS so there might be additional MacOS specific rendering issues.

Lythenas:
@sebnow can you check again if that fixed (some of) the graphical artifacts. It would also be helpful to know if you only have these issues on MacOS or on your other machine too.

sebnow:
It did appear to help. It's much harder to get artifacts, but some still appear.
Screen Shot 2019-09-19 at 10 44 23

> > Ignore the font rendering issues - I haven't played around with the options yet.
>
> Don’t think any options can change those font issues (correct me if I’m wrong); they shouldn’t be happening, so I didn’t ignore them in my above response.

I was specifically referring to the lack of some characters, as I didn't have `guifont` set and my statusline requires some glyphs not present in the default font.

> See #85. Not sure if that is the problem, but it looks like it to me, so worth a shot.

I don't have time to play around with it now, but it does appear to be the same issue.
Screen Shot 2019-09-19 at 10 55 30

Lythenas:
How are we going to continue?

It would be good if more people (especially with MacOS) could test it but on all my machines (Arch with pango 1.43) it works as expected and without any graphical issues.

smolck:
Now we need someone to test this with pango 1.43 on MacOS. Either we find (or wait for) someone who can do that or maybe @sebnow could try that when he gets the time?

sebnow:
Sorry, I should have annotated it. It's easy to miss.
65228316-88824b00-daca-11e9-9d7c-52bd88820cfa

smolck:
> Considering the font rendering issues, with some lines overlapping, it may be related.

It sure does look that way to me. @sebnow Would you be able to try the diff mentioned here and see if it removes the artifacts: https://github.com/vhakulinen/gnvim/issues/85#issuecomment-529230236? It isn’t a perfect fix, but it’ll remove the pango-related graphical issues enough to where you can see the issues specific to this branch (if there are any).

Issue #43, opened 2019-05-10
[Bug] Netrw crashes gnvim


elihunter173:
When I open gnvim from the terminal and go to Netrw by executing `:Explore` (or by any other method), it acts as expected and takes me to Netrw.

However, when I open gnvim outside of the terminal (namely executing it from Budgie's menu), whenever I attempt to open Netrw, gnvim immediately crashes with no error message or warning.

Edit: Fixed typo

elihunter173:
Actually, it is now occurring even when running gnvim from the terminal.

### Error Message
```
thread 'main' panicked at 'byte index 2 is out of bounds of ``', src/libcore/str/mod.rs:2010:9
stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: <unknown>
  12: <unknown>
  13: <unknown>
  14: g_main_context_dispatch
  15: <unknown>
  16: g_main_context_iteration
  17: g_application_run
  18: <unknown>
  19: <unknown>
  20: <unknown>
  21: <unknown>
  22: <unknown>
  23: <unknown>
  24: <unknown>
  25: __libc_start_main
  26: <unknown>
```

badosu:
Try master, should be fixed:  https://github.com/vhakulinen/gnvim/issues/24

elihunter173:
Sadly I run into the same issue where I works when launching from the terminal but not outside it.

badosu:
One reason could be having two versions of gnvim installed, the one launched outside of terminal being outdated.

elihunter173:
That would make sense. However, whenever I run `sudo make uninstall`, I can no longer start up gnvim from either place. When I do `sudo make install`, then both begin working again and the same issue persists.

vhakulinen:
Can you do `RUST_BACKTRACE=1 cargo run`? `cargo run` runs the debug build which would give more information.

vhakulinen:
I managed to reproduce this issue and it looks like this is related to #29, which in turn is caused by a bug in neovim. Related neovim issue: https://github.com/neovim/neovim/issues/10000.

elihunter173:
[...]

PR #123, opened 2019-11-23
Check whether working with valid index

matthiasbeyer:
This patch is a first draft for a bugfix where gnvim crashes when using
:Explore in $HOME (and also sometimes in non-$HOME).

It adds a check whether the value used to index
`self.content.split_at()` is within the buffer and at a valid unicode
boundary for the `String`.

---

Followup to the discussion in #43

This is a **draft**. I am not an expert on this unicode-stuff and neither am I for this codebase, so maybe we want to iterate on this patch a bit?

### DIFF
--- a/src/ui/cmdline.rs
+++ b/src/ui/cmdline.rs
@@ -401,13 +401,17 @@ impl CmdlineInput {
         let buffer = self.textview.get_buffer().unwrap();
         let mut iter = buffer.get_start_iter();
 
-        let pos = self.content.split_at(self.cursor_pos).0.chars().count();
+        if self.content.as_bytes().len() > self.cursor_pos
+            && self.content.is_char_boundary(self.cursor_pos)
+        {
+            let pos = self.content.split_at(self.cursor_pos).0.chars().count();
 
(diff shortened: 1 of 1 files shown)


smolck:
Is this necessary? (This doesn’t happen in the newer Neovim versions.)

matthiasbeyer:
But in older ones and some distros do not update that fast.

I do not say it is a solution, but it can be a base for one.

src/ui/cmdline.rs
```diff
@@ -401,13 +401,17 @@ impl CmdlineInput {
         let buffer = self.textview.get_buffer().unwrap();
         let mut iter = buffer.get_start_iter();
 
-        let pos = self.content.split_at(self.cursor_pos).0.chars().count();
+        if self.content.as_bytes().len() > self.cursor_pos
```
vhakulinen: The `is_char_boundary` ought to be enough.

vhakulinen:
As far as I've understood, this doesn't happen on the latest nvim release which is a requirement for gnvim. Rejecting based on that since I don't really want to watch out for bugs in nvim's older versions in gnvim's code.

PR #124, opened 2019-12-01
[WIP] Fix #85

### LINKED ISSUE #85: Font rendering issues with pango 1.44

smolck:
This is fairly close to working, but it has visual artifacts (and the font looks a bit blurred to me, but maybe not for everyone?):

image

Interestingly, they disappear after opening the cmdline (so after typing `:`):

image

The line that's causing this "reset" of the artifacts is here (see https://github.com/vhakulinen/gnvim/issues/85#issuecomment-560044135): https://github.com/vhakulinen/gnvim/blob/65846f45020b338fedd932b058c9e8036dadd976/src/ui/cmdline.rs#L575

But I have no idea why calling `show_all` on a `gtk::Fixed` removes the artifacts. Any ideas?

The artifacts appear to be coming from a lack of subpixel positioning in GNvim, but I could be wrong. Note that when I `ceil` (or `floor`) the `width` value of the `CellMetrics` the artifacts disappear, but the cursor is no longer positioned correctly. (All this still doesn't explain the "reset" when opening the cmdline though.)

cc: @vhakulinen @last-partizan

### DIFF
--- a/src/ui/grid/context.rs
+++ b/src/ui/grid/context.rs
@@ -155,7 +155,7 @@ impl Context {
     }
 
     /// Returns x, y, width and height for current cursor location.
-    pub fn get_cursor_rect(&self) -> (f64, f64, f64, f64) {
+    pub fn get_cursor_rect(&self) -> (i32, i32, i32, i32) {
         let double_width = self
             .rows
             .get(self.cursor.0 as usize)
@@ -172,14 +172,14 @@ impl Context {
             self.cursor.1 as f64,
         );
         (
-            x,
-            y,
+            x.ceil() as i32,
+            y.ceil() as i32,
             if double_width {
-                cm.width * 2.0
+                (cm.width * 2.0).ceil() as i32
             } else {
-                cm.width
+                (cm.width).ceil() as i32
             },
-            cm.height,
+            cm.height.ceil() as i32,
         )
     }
 }
--- a/src/ui/grid/context.rs
+++ b/src/ui/grid/context.rs
@@ -204,10 +204,13 @@ impl CellMetrics {
             .get_metrics(Some(&self.font.as_pango_font()), None)
             .unwrap();
         let extra = self.line_space as f64 / 2.0;
-        self.ascent = fm.get_ascent() as f64 / pango::SCALE as f64 + extra;
-        self.decent = fm.get_descent() as f64 / pango::SCALE as f64 + extra;
-        self.height = self.ascent + self.decent;
-        self.width = (fm.get_approximate_digit_width() / pango::SCALE) as f64;
+        self.ascent =
+            (fm.get_ascent() as f64 / pango::SCALE as f64 + extra).ceil();
+        self.decent =
+            (fm.get_descent() as f64 / pango::SCALE as f64 + extra).ceil();
+        self.height = self.ascent + self.decent + self.line_space as f64;
+        self.width =
+            fm.get_approximate_char_width() as f64 / pango::SCALE as f64;
 
         self.underline_position =
             fm.get_underline_position() as f64 / pango::SCALE as f64 - extra;
--- a/src/ui/grid/grid.rs
+++ b/src/ui/grid/grid.rs
@@ -511,6 +511,7 @@ fn drawingarea_draw(cr: &cairo::Context, ctx: &mut Context) {
     // If we're not "busy", draw the cursor.
     if !ctx.busy {
         let (x, y, w, h) = ctx.get_cursor_rect();
+        let (x, y, w, h) = (x.ceil(), y.ceil(), w.ceil(), h.ceil());
 
         cr.save();
         cr.rectangle(x, y, w * ctx.cursor_cell_percentage, h);
--- a/src/ui/grid/grid.rs
+++ b/src/ui/grid/grid.rs
@@ -330,30 +330,20 @@ impl Grid {
 
         // Clear old cursor position.
         let (x, y, w, h) = ctx.get_cursor_rect();
-        ctx.queue_draw_area.push((
-            x.ceil() as i32,
-            y.ceil() as i32,
-            w.ceil() as i32,
-            h.ceil() as i32,
-        ));
+        ctx.queue_draw_area.push((x, y, w, h));
         ctx.cursor.0 = row;
         ctx.cursor.1 = col;
 
         // Mark the new cursor position to be drawn.
         let (x, y, w, h) = ctx.get_cursor_rect();
-        ctx.queue_draw_area.push((
-            x.ceil() as i32,
-            y.ceil() as i32,
-            w.ceil() as i32,
-            h.ceil() as i32,
-        ));
+        ctx.queue_draw_area.push((x, y, w, h));
 
         if let Some(ref im_context) = self.im_context {
             let rect = gdk::Rectangle {
-                x: x as i32,
-                y: y as i32,
-                width: w as i32,
-                height: h as i32,
+                x: x,
+                y: y,
+                width: w,
+                height: h,
             };
             im_context.set_cursor_location(&rect);
         }
@@ -455,8 +445,7 @@ impl Grid {
         // Don't use the ctx.queue_draw_area, because those draws will only
         // happen once nvim sends 'flush' event. This draw needs to happen
         // on each tick so the cursor blinks.
-        self.da
-            .queue_draw_area(x as i32, y as i32, w as i32, h as i32);
+        self.da.queue_draw_area(x, y, w, h);
     }
 
     /// Set a new font and line space. This will likely change the cell metrics.
@@ -511,13 +500,17 @@ fn drawingarea_draw(cr: &cairo::Context, ctx: &mut Context) {
     // If we're not "busy", draw the cursor.
     if !ctx.busy {
         let (x, y, w, h) = ctx.get_cursor_rect();
-        let (x, y, w, h) = (x.ceil(), y.ceil(), w.ceil(), h.ceil());
 
         cr.save();
-        cr.rectangle(x, y, w * ctx.cursor_cell_percentage, h);
+        cr.rectangle(
+            x as f64,
+            y as f64,
+            w as f64 * ctx.cursor_cell_percentage,
+            h as f64,
+        );
         let surface = ctx.cursor_context.get_target();
         surface.flush();
-        cr.set_source_surface(&surface, x, y);
+        cr.set_source_surface(&surface, x as f64, y as f64);
         cr.fill();
         cr.restore();
     }
--- a/src/ui/grid/grid.rs
+++ b/src/ui/grid/grid.rs
@@ -330,15 +330,23 @@ impl Grid {
 
         // Clear old cursor position.
         let (x, y, w, h) = ctx.get_cursor_rect();
-        ctx.queue_draw_area
-            .push((x as i32, y as i32, w as i32, h as i32));
+        ctx.queue_draw_area.push((
+            x.ceil() as i32,
+            y.ceil() as i32,
+            w.ceil() as i32,
+            h.ceil() as i32,
+        ));
         ctx.cursor.0 = row;
         ctx.cursor.1 = col;
 
         // Mark the new cursor position to be drawn.
         let (x, y, w, h) = ctx.get_cursor_rect();
-        ctx.queue_draw_area
-            .push((x as i32, y as i32, w as i32, h as i32));
+        ctx.queue_draw_area.push((
+            x.ceil() as i32,
+            y.ceil() as i32,
+            w.ceil() as i32,
+            h.ceil() as i32,
+        ));
 
         if let Some(ref im_context) = self.im_context {
             let rect = gdk::Rectangle {
--- a/src/ui/grid/render.rs
+++ b/src/ui/grid/render.rs
@@ -159,7 +159,12 @@ fn put_segments(
         let text = seg.leaf.text();
         render_text(cr, pango_context, cm, &hl, hl_defs, &text, x, y, w, h);
 
-        queue_draw_area.push((x as i32, y as i32, w as i32, h as i32));
+        queue_draw_area.push((
+            x.ceil() as i32,
+            y.ceil() as i32,
+            w.ceil() as i32,
+            h.ceil() as i32,
+        ));
     }
 }
 
@@ -304,8 +309,12 @@ pub fn scroll(ctx: &mut Context, hl_defs: &HlDefs, reg: [u64; 4], count: i64) {
     cr.set_operator(cairo::Operator::Source);
     cr.rectangle(x1, y1, w, h);
     cr.fill();
-    ctx.queue_draw_area
-        .push((x1 as i32, y1 as i32, w as i32, h as i32));
+    ctx.queue_draw_area.push((
+        x1.ceil() as i32,
+        y1.ceil() as i32,
+        w.ceil() as i32,
+        h.ceil() as i32,
+    ));
 
     // Clear the area that is left "dirty".
     let (x1, y1, x2, y2) = get_rect(
@@ -321,8 +330,12 @@ pub fn scroll(ctx: &mut Context, hl_defs: &HlDefs, reg: [u64; 4], count: i64) {
     cr.rectangle(x1, y1, x2 - x1, y2 - y1);
     cr.set_source_rgb(bg.r, bg.g, bg.b);
     cr.fill();
-    ctx.queue_draw_area
-        .push((x1 as i32, y1 as i32, w as i32, h as i32));
+    ctx.queue_draw_area.push((
+        x1.ceil() as i32,
+        y1.ceil() as i32,
+        w.ceil() as i32,
+        h.ceil() as i32,
+    ));
 
     cr.restore();
 }


last-partizan:
With this fixes blurring on scrool is gone, and i see almost no artifacts.

Hovewer, artifacts appear when moving cursor from end to start in this line:
Screenshot from 2019-12-01 10-31-20

It looks like new bug, with master branch it works without artifacts.

vhakulinen:
Some code in [render.rs](https://github.com/vhakulinen/gnvim/blob/65846f45020b338fedd932b058c9e8036dadd976/src/ui/grid/render.rs) is casting `f64` to `i32`. My guess is that that is causes those artifacts with cursor movement.

smolck:
> Some code in [render.rs](https://github.com/vhakulinen/gnvim/blob/65846f45020b338fedd932b058c9e8036dadd976/src/ui/grid/render.rs) is casting `f64` to `i32`. My guess is that that is causes those artifacts with cursor movement.

Would you mind pointing out the code you're speaking of here?

vhakulinen:
This particular case, here: https://github.com/vhakulinen/gnvim/blob/65846f45020b338fedd932b058c9e8036dadd976/src/ui/grid/render.rs#L115-L136

vhakulinen:
Sorry, the cursor stuff was in grid.rs: https://github.com/vhakulinen/gnvim/blob/65846f45020b338fedd932b058c9e8036dadd976/src/ui/grid/grid.rs#L327-L352

smolck:
@vhakulinen Is there an alternative to `gdk::Rectangle` we could use that takes floats? There isn't really a way to avoid the casting to `i32` without changing that because a `gdk::Rectangle` takes `i32`s.

vhakulinen:
Probably not, but I'm not sure of its significance for the `im_context`. The ones for `queue_draw_area` should be rounded up and cast to `i32`.

smolck:
> The ones for `queue_draw_area` should be rounded up and cast to `i32`.

Thanks! I've pushed your changes, and that's removed most of the artifacts (!). The only ones I've noticed happen when opening a file with vim-startify, so I'm not sure what's causing that. Otherwise it works great for me!

@last-partizan Could you test this branch again and see if it works for you?

last-partizan:
@smolck Looks like it still has artifacts after cursor movement for me.

smolck:
> @smolck Looks like it still has artifacts after cursor movement for me.

Interesting. I'm not getting artifacts on cursor movement, do they look the same (and occur in the same way) as those you saw in https://github.com/vhakulinen/gnvim/pull/124#issuecomment-560076267?

smolck:
@last-partizan I've pushed a commit ceiling the values to `queue_draw_area` in context.rs, and it's appears to have removed all of the artifacts for me. Would you mind trying again with the latest commit?

last-partizan:
Still don't work on my laptop, wait a minute i will check on desktop.

smolck:
> Still don't work on my laptop, wait a minute i will check on desktop.

Okay, thanks! What version of `pango` are you using on your laptop? I'm using version `1:1.44.7-1`.

last-partizan:
@smolck same version.

Doesn't work on desktop either.

last-partizan:
https://imgur.com/0cF0Wow

here is how it looks

last-partizan:
I was thinking, maybe it just on wayland, but no. Tried under xorg, same result.

smolck:
Hmm, that's strange. I don't get any artifacts now....

@liyiheng In https://github.com/vhakulinen/gnvim/issues/85#issuecomment-532198604 you mentioned you had issues with pango and GNvim. Can you test this branch and see if it fixes the problems you've had?

smolck:
@last-partizan I guess I wasn't testing things enough, I'm getting the artifacts your seeing too. I'm going to see if I can fix/improve that. Thanks for the info.

smolck:
It's interesting how the artifacts appear almost exclusively when moving the cursor to the right, and how the artifacts disappear as I move the cursor back to the left (so that the cursor goes over the artifacts).

smolck:
Also interesting is that if I do `:set guifont=Iosevka\ Extrabold:h14.5` so that I have a floating-point font size, not as many artifacts appear when moving the cursor, or at least the artifacts are more spread out.

liyiheng:
@smolck 
pango 1:1.44.7-1
Commit 001d4ffa3f88bf26e4eab407d9cfa11d309aa4d5

I tested `H`,`J`,`K`,`L`.
`H`,`J`,`K` are fine, but `L`:
image

smolck:
> I tested `H`,`J`,`K`,`L`.
> `H`,`J`,`K` are fine, but `L` . . .

@vhakulinen Any idea what could be causing the artifacts only when moving the cursor with `L`?

vhakulinen:
Hmh, could be the `Grid::tick` function. It should also ceil the values before casting to i32.

smolck:
@vhakulinen Still have artifacts when ceiling the values pushed to `queue_draw_area` at the end of `Grid::tick`, both with and without `:set guicursor += a:blinkon333`.

vhakulinen:
Maybe you need to floor the `x` values.

smolck:
I was able to remove the cursor-caused artifacts on my machine by ceiling the values used for drawing the cursor in `drawingarea_draw()`. I think the artifacts were caused by the values for the cursor's rectangle's size being inconsistent between `queue_draw_area` and the `drawingarea_draw()` function. So, after ceiling the cursor rectangle values used in `drawingarea_draw` the artifacts no longer occur, at least for me.

@last-partizan @liyiheng Can you guys try this branch once more with the latest commit I just pushed? The cursor should no longer cause any artifacts. Thank you for testing this PR btw, doing so helps out a lot.

@vhakulinen Do you think this PR is now good for merge (assuming there are no more problems for last-partizan or liyiheng), or is there anything we should/can clean up or polish? 

I'd like to at least make the `ceil`ing of the values passed to `queue_draw_area` happen in one place, by creating a function to do so or adapting existing functions to return `ceil`ed `f64`s as `i32`s. That way, we (1) don't repeat ourselves and (2) make it easier to maintain consistency throughout the code.

liyiheng:
Commit 150b1a7c87097f00a37898fd9c268c63b36bbd39 
 @smolck Everything is OK on my laptop

smolck:
@vhakulinen I've pushed https://github.com/vhakulinen/gnvim/pull/124/commits/b383efd2e4fdecb88c2caa76552aea406a495c91, which I think is a good first step for making this code cleaner, but I think `render.rs` still needs some work. I've removed the `[WIP]` label from the PR in any case, since this is now ready for review and polish, and not really a "work-in-progress" anymore.

last-partizan:
I can confirm, artifacts under cursor are gone.

But one little is still present is statusbar: 
Screenshot from 2019-12-15 12-26-22

There is thin vertical lines, and they are gone after area is repainted (for example by changing vim mode to edit and back)
Screenshot from 2019-12-15 12-28-23

smolck:
Yeah, I thought I noticed some artifacts in the statusbar. Thanks for the info, I’ll see what I can come up with.

smolck:
Interesting. So if I change this:

https://github.com/vhakulinen/gnvim/blob/b383efd2e4fdecb88c2caa76552aea406a495c91/src/ui/grid/context.rs#L213

into this:

```rust
fm.get_approximate_digit_width() as f64 / pango::SCALE as f64; 
```

Then the problem with the lines in the statusline is fixed. But then the text inside the editor is then too close together in some places, like in this example:

image

Any ideas?

cc: @vhakulinen

smolck:
@vhakulinen It feels wrong and hacky to just `ceil` values everywhere, rather than find a way to properly support floating-point `CellMetrics`; I think we should try to fix the issues without `ceil`ing values. 

Here's a starting diff for this, based off of master:
```diff
diff --git a/src/ui/grid/context.rs b/src/ui/grid/context.rs
index e7ca63b..8844577 100644
--- a/src/ui/grid/context.rs
+++ b/src/ui/grid/context.rs
@@ -43,7 +43,7 @@ pub struct Context {
     pub active: bool,
 
     /// Areas to call queue_draw_area on the drawing area on flush.
-    pub queue_draw_area: Vec<(i32, i32, i32, i32)>,
+    pub queue_draw_area: Vec<(f64, f64, f64, f64)>,
 }
 
 impl Context {
@@ -207,7 +207,7 @@ impl CellMetrics {
         self.ascent = fm.get_ascent() as f64 / pango::SCALE as f64 + extra;
         self.decent = fm.get_descent() as f64 / pango::SCALE as f64 + extra;
         self.height = self.ascent + self.decent;
-        self.width = (fm.get_approximate_digit_width() / pango::SCALE) as f64;
+        self.width = fm.get_approximate_char_width() as f64 / pango::SCALE as f64;
 
         self.underline_position =
             fm.get_underline_position() as f64 / pango::SCALE as f64 - extra;
diff --git a/src/ui/grid/grid.rs b/src/ui/grid/grid.rs
index a9e2b04..65f9938 100644
--- a/src/ui/grid/grid.rs
+++ b/src/ui/grid/grid.rs
@@ -137,7 +137,7 @@ impl Grid {
         ctx.cursor_color = hl.foreground.unwrap_or(hl_defs.default_fg);
 
         while let Some(area) = ctx.queue_draw_area.pop() {
-            self.da.queue_draw_area(area.0, area.1, area.2, area.3);
+            self.da.queue_draw_area(area.0 as i32, area.1 as i32, area.2 as i32, area.3 as i32);
         }
     }
 
@@ -331,14 +331,14 @@ impl Grid {
         // Clear old cursor position.
         let (x, y, w, h) = ctx.get_cursor_rect();
         ctx.queue_draw_area
-            .push((x as i32, y as i32, w as i32, h as i32));
+            .push((x, y, w, h));
         ctx.cursor.0 = row;
         ctx.cursor.1 = col;
 
         // Mark the new cursor position to be drawn.
         let (x, y, w, h) = ctx.get_cursor_rect();
         ctx.queue_draw_area
-            .push((x as i32, y as i32, w as i32, h as i32));
+            .push((x, y, w, h));
 
         if let Some(ref im_context) = self.im_context {
             let rect = gdk::Rectangle {
diff --git a/src/ui/grid/render.rs b/src/ui/grid/render.rs
index c4359b4..9607b42 100644
--- a/src/ui/grid/render.rs
+++ b/src/ui/grid/render.rs
@@ -139,7 +139,7 @@ pub fn cursor_cell(
 fn put_segments(
     cr: &cairo::Context,
     pango_context: &pango::Context,
-    queue_draw_area: &mut Vec<(i32, i32, i32, i32)>,
+    queue_draw_area: &mut Vec<(f64, f64, f64, f64)>,
     cm: &CellMetrics,
     hl_defs: &HlDefs,
     segments: Vec<Segment>,
@@ -159,7 +159,7 @@ fn put_segments(
         let text = seg.leaf.text();
         render_text(cr, pango_context, cm, &hl, hl_defs, &text, x, y, w, h);
 
-        queue_draw_area.push((x as i32, y as i32, w as i32, h as i32));
+        queue_draw_area.push((x, y, w, h));
     }
 }
 
@@ -227,7 +227,7 @@ pub fn clear(da: &DrawingArea, ctx: &mut Context, hl_defs: &HlDefs) {
     cr.fill();
     cr.restore();
 
-    ctx.queue_draw_area.push((0, 0, w, h));
+    ctx.queue_draw_area.push((0.0, 0.0, w as f64, h as f64));
 }
 
 /// Scrolls contents in `ctx.cairo_context` and `ctx.rows`, based on `reg`.
@@ -305,7 +305,7 @@ pub fn scroll(ctx: &mut Context, hl_defs: &HlDefs, reg: [u64; 4], count: i64) {
     cr.rectangle(x1, y1, w, h);
     cr.fill();
     ctx.queue_draw_area
-        .push((x1 as i32, y1 as i32, w as i32, h as i32));
+        .push((x1, y1, w, h));
 
     // Clear the area that is left "dirty".
     let (x1, y1, x2, y2) = get_rect(
@@ -322,7 +322,7 @@ pub fn scroll(ctx: &mut Context, hl_defs: &HlDefs, reg: [u64; 4], count: i64) {
     cr.set_source_rgb(bg.r, bg.g, bg.b);
     cr.fill();
     ctx.queue_draw_area
-        .push((x1 as i32, y1 as i32, w as i32, h as i32));
+        .push((x1, y1, w, h));
 
     cr.restore();
 }
```

Changing `self.width = (fm.get_approximate_digit_width() / pango::SCALE) as f64;` to `self.width = fm.get_approximate_char_width() as f64 / pango::SCALE as f64;` fixed the issues with text being too close together and the cursor not being in the correct position, but there are still artifacts and blurry scrolling.

Blurry scrolling means the `scroll()` function in `render.rs` will probably need some tweaking, although I'm not sure what tweaking is needed (I'm unfamiliar with how scrolling works).

Note that while we *were* working with `f64`s before pango 1.44, the values ended in `.0`, whereas now they do not (see https://github.com/vhakulinen/gnvim/issues/85#issuecomment-527195132). I think this is what's causing the blurry scrolling, but I'm not sure of the fix. It does appear to be related to the font height, because as @last-partizan pointed out in #85 `ceil`ing the height in `CellMetrics` fixes the blurry scrolling.

I think making `queue_draw_area` in `Context` a `Vec<(f64, f64, f64, f64)>` instead of a `Vec<(i32, i32, i32, i32)>` will help with floating-point support, so the above diff does that too.

vhakulinen:
I'll have a look at this once I find the time.

vhakulinen:
Anyone having probmels with pango 1.44, can you try https://github.com/vhakulinen/gnvim/tree/fix-pango144?

elihunter173:
The issues with fuzzy text when scrolling and the cursor not moving the width of the characters seems to be resolved for me on pango 1:1.44.7-1 (Arch Linux).

However, now the view port seems to think that it can fit more characters than it can. In the image below, you can see my status bar isn't shown and there are 14 a's off the screen.

image

Also, I couldn't get a good screenshot, but bold text seems to have a slightly different width than non-bold text. Causing a very minor jitter when I was searching things with fzf.vim.

Side note, I really appreciate the work you're doing, and I look forward to be able to use gnvim as my daily driver again!

elihunter173:
Fixed the issue (as far as I can tell at least). It was just getting some pretty bad rounding error because of converting width (and height) from a float to an integer and then doing division with integers rather than doing division with floats and then converting to an integer.

I haven't uploaded the commit anywhere, but here's the diff

```diff
diff --git a/src/ui/grid/grid.rs b/src/ui/grid/grid.rs
index 2c8e291..a3b176e 100644
--- a/src/ui/grid/grid.rs
+++ b/src/ui/grid/grid.rs
@@ -364,17 +364,17 @@ impl Grid {
         }
     }

-    /// Calcualtes the current size of the grid.
+    /// Calculates the current size of the grid in cells.
     pub fn calc_size(&self) -> (i64, i64) {
         let mut ctx = self.context.borrow_mut();
         let ctx = ctx.as_mut().unwrap();

         let w = self.da.get_allocated_width();
         let h = self.da.get_allocated_height();
-        let cols = (w / ctx.cell_metrics.width as i32) as i64;
-        let rows = (h / ctx.cell_metrics.height as i32) as i64;
+        let cols = w as f64 / ctx.cell_metrics.width;
+        let rows = h as f64 / ctx.cell_metrics.height;

-        (cols, rows)
+        (cols as i64, rows as i64)
     }

     pub fn resize(&self, width: u64, height: u64) {
```

smolck:
@vhakulinen Your branch fixes the issues for me (!), except for the problem @elihunter173 stated above. Unfortunately, his patch *does not* fix the statusline problem for me.

However, the issue with rounding values is that things aren't quite what they should be. Most notably is the status bar. With pango 1.44 and the fix-pango144 branch (**with** elihunter's patch applied) the (entire) bottom of GNvim looks like this (note the gap between the statusbar and the bottom of the window):

image

On master branch however, GNvim looks like this (notice the statusbar is almost touching the bottom of the window):

image

This is not necessarily a deal-breaker for me, but I prefer the look of the statusline without that gap (note that this probably won't matter **without** the external cmdline enabled since a gap is necessary for the cmdline, but **with** the external cmdline enabled there's no reason for there to be such a gap).

elihunter173:
After more testing, I can confirm my patch doesn't fix the bug as @smolck mentioned.

smolck:
Maybe we should wait on a new release for the Rust bindings to Pango? I wonder if that could have an effect on whether or not sub-pixel positioning is enabled; from my debugging in #85, I found that even after they changed sub-pixel positioning’s default to off upstream in pango 1.44, the `CellMetrics` (defined in `src/ui/grid/context.rs` iirc) still had different numbers than those before pango 1.44, and those new numbers had several decimal places of precision they didn’t previously have (e.g., using arbitrary example numbers,`10.76755342` for the cell height in pango 1.44 versus `10.0` in pango 1.43).

See https://github.com/gtk-rs/pango/issues/165. At the very least, updated pango bindings would give us more functions to use to try to replicate GNvim’s pre-pango 1.44 behavior.

We can also explore the commit history and source code from other applications using pango and see what those apps did to fix the pango issues on their end.

smolck:
Note also this could at least partly be an upstream issue, see https://github.com/harfbuzz/harfbuzz/issues/1892#issuecomment-560097572 (and also the whole thread).

vhakulinen:
Merged in https://github.com/vhakulinen/gnvim/commit/13bd8b0590f43881661ca02cb81bf6a5e65e02eb

PR #126, opened 2020-02-07
Makefile: install on FreeBSD

ggouss:
somehow install utility works differently on FreeBSD. This commit makes build && install process seamless on FreeBSD machine

I do not have linux box with UI right now, so cannot test with new Makefile on linux. If somebody can - it would be great


Jberlinsky:
FWIW, This also affects OS X users

PR #143, opened 2020-08-05
Add flag to prefer dark theme

### LINKED ISSUE #142: Doesn't seem to respect GTK_THEME environment variable
I'd love it if I could change gnvim's window border using `GTK_THEME=Adwaita:dark gnvim`, but this doesn't seem to have any effect.

From the [Running GTK+ Applications page](https://developer.gnome.org/gtk3/stable/gtk-running.html):

> `GTK_THEME`.  If set, makes GTK+ use the named theme instead of the theme that is specified by the gtk-theme-name setting. This is intended mainly for easy debugging of theme issues. It is also possible to specify a theme variant to load, by appending the variant name with a colon, like this: `GTK_THEME=Adwaita:dark`. 

Thanks for the great app!

jacobmischka:
Fixes #142, for me at least.

Still would probably be nice to actually support that env var, but this is what I wanted it for anyway.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -94,6 +94,10 @@ struct Options {
     #[structopt(long = "disable-ext-tabline")]
     disable_ext_tabline: bool,
 
+	/// Enables dark theme
+	#[structopt(long = "prefer-dark-theme")]
+	prefer_dark_theme: bool,
+
     /// Geometry of the window in widthxheight form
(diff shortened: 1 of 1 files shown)


vhakulinen:
Thanks! I made a small change to the flag name (added `gtk` prefix), rebased and merged this to master. Apparently, because of the rebase, github can't track that these changes (commits) were merged.

Ref: https://github.com/vhakulinen/gnvim/commit/336edb667b4235b5b7bbbf39de166c6a2e3fd0ce

PR #155, opened 2020-11-01
Added support for drag and drop (files and text)

Amjad50:
This commit tries to replicate how GVim handles drop events for files
and text.

added:
- When text is dropped, it will pasted to the current buffer and cursor (nvim_paste).
- When files are dropped when `cmdline`  is closed, they will be opened using `:edit` or `:split` (`split` if the current buffer is modified).
- When files are dropped when `cmdline` is opened, the names will be added into the `cmdline` buffer

> Note: not sure why `Cargo.lock` changed to much, only added `urlencoding`. Its just that the checksum values changed place.

What do you think @vhakulinen ?

### DIFF
--- a/src/ui/ui.rs
+++ b/src/ui/ui.rs
@@ -1,5 +1,6 @@
 use std::cell::RefCell;
 use std::collections::HashMap;
+use std::path::Path;
 use std::rc::Rc;
 
 use gtk::prelude::*;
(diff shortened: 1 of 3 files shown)


src/ui/ui.rs
```diff
@@ -390,3 +437,45 @@ fn event_to_nvim_input(e: &gdk::EventKey) -> Option<String> {
...
+    nvim: GioNeovim,
+    data: &str,
+) -> Result<bool, Box<nvim_rs::error::CallError>> {
+    // since [`nvim_rs`] does not provide `paste`, it is being called manually
+    // here
+    nvim.call("nvim_paste", nvim_rs::call_args![data, false, -1])
+        .await??
+        .try_unpack()
+        .map_err(|v| Box::new(nvim_rs::error::CallError::WrongValueType(v)))
+}
```
Amjad50, line 481: I think this can be made into `nvim_rs` as a PR

src/ui/ui.rs
```diff
@@ -390,3 +435,45 @@ fn event_to_nvim_input(e: &gdk::EventKey) -> Option<String> {
...
+                        // only respond to
+                        // "Vim(edit):E37: No write since last change (add ! to override)"
+                        if s.starts_with("Vim(edit):E37") {
+                            // ignore errors
+                            nvim.command(&format!("sp {}", filename))
+                                .await
+                                .ok();
+                        }
+                    }
+                }
```
Amjad50, line 463: In the GVim code, it determines if it needs to `split` or `edit` by checking if the current buffer has `changed`, but I could not find a way to check for that, even `nvim_get_option('modified')` does not provide that

vhakulinen:
Hi. I'm a bit busy with life atm and don't have the time/energy to work on gnvim currently. Drag & drop is definitely a feature that gnvim should have and I _really_ appreciate the work you've done here. But this will have to wait from my site for a while.

Again, I really appreciate the work you've done here.

fidelski:
Is there any plan to work on this? I would appreciate this feature very much.

vhakulinen:
Deprecated.

PR #175, opened 2021-07-03
Implement smooth scrolling support

### LINKED ISSUE #70: No touchpad (libinput) scroll on sway WM (wayland)
Is there a setting to enable scrolling with touchpad? I can select with the mouse, but not scroll. Thanks!

jeanguyomarch:
Scrolling with the touchpad had no effect. It seemed scrolling events
were purely ignored. As explained in #70, this is due to smooth
scrolling. The problem with enabling smooth scrolling is that it
generates a lot of scrolling events, resulting in overly fast scroll
without further processing.

The solution here consists in hooking on the "start scroll" event to
reset a scrolling counter (this requires to bump the GTK version). When
the scrolling is hapening, this counter is incremented with the
scrolling deltas.  When it reaches -1.0 or +1.0 (which is the value that
"normal scrolling" sends), we trigger a scrolling event (down or up,
depending on the sign) and reset the counter.

Overall, it enables scrolling in gnvim with the touchpad.

This probably fixes #70.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ rmpv = "0.4"
 
 [dependencies.gtk]
 version = "0.9"
-features = ["v3_18"]
+features = ["v3_24"]
 
 [dependencies.webkit2gtk]
 version = "0.10"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ rmpv = "0.4"
 
 [dependencies.gtk]
 version = "0.9"
-features = ["v3_24"]
+features = ["v3_18"]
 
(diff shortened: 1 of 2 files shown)


vhakulinen:
Nice, thanks for the PR! I'll to take a closer look and test this within a week or so.

vhakulinen:
I dropped the gtk version requirement back to 3.18 so we don't change the requirements for gnvim. This meant that the EventControllerScroll usage had to be removed, but scrolling still worked on wayland + touchbad combo. Only down side is that the delta value isn't rested anymore on scroll start.

PR #183, opened 2021-11-05
version is 0.1.6

maage:
Update version number to match git tag to reduce confusion.


vhakulinen:
Deprecated.

PR #187, opened 2021-12-08
Remove semicolons in macros with expressions

agraven:
This fixes a compiler warning which was introduced in a recent version
of the rust compiler. The warning will eventually be elevated to an
error.


vhakulinen:
Deprecated.

PR #194, opened 2022-10-19
Fix typos

kianmeng:
Found via `typos --format brief`


vhakulinen:
Deprecated.

PR #196, opened 2022-12-08
Update gtk4 install hint

tjwallace:
I could not find a reference to gtk-4-dev in any Debian based repositories.


vhakulinen:
Thanks!

PR #211, opened 2023-05-22
RFC / DRAFT: Add release-debug, optimized, and optimized-debug build profiles.

linuxtim:
I've done a bit of rust development on bare metal embedded systems, where I usually tweak the default cargo release build profile for minimum code size.  I thought I would try the same on gnvim, and it reduced the generated code size by 45% whilst only adding 10% to the compile time.   I thought this was probably worth having, so here's a PR which introduces this build profile, and changes the Makefile to use it by default.

There are also a couple of other profiles which I've used in the past for debug builds for troubleshooting, which I thought could provide a straightforward way to help users debug issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,3 +4,30 @@ members = [
     "lib/nvim-rs",
     "lib/gio-compat",
 ]
+
+# https://doc.rust-lang.org/cargo/reference/profiles.html
+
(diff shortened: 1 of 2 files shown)


vhakulinen:
Thanks. This reminded me about some dev build optimizations that the bevy game engine recommends, so I added those too. It made the dev build go from "not usable" to "somewhat usable for testing".

PR #215, opened 2023-09-01
ui: add wrap parameter for grid_line event (resolves #214)

### LINKED ISSUE #214: Black screen on nvim 0.10.x (panic)
Running gnvim on Pop!_OS 22.04 LTS with nvim:
```
NVIM v0.10.0-dev-995+gb2c3e9c72
Build type: RelWithDebInfo
LuaJIT 2.1.1693350652
```

Getting the following error:
```
thread 'main' panicked at 'failed to decode redraw notification: Syntax("error while decoding value: invalid length 5, expected fewer elements in array")', ui/src/components/appwindow/imp.rs:84:26
```

Results in a near-blank black screen loading.

Digging deeper, this seems to be a deserialization issue with one of the UIEvents. specifically the `grid_line` event in `lib/nvim-rs/src/types/gen.rs:574` where it tries to deserialize the event to a `GridLine` struct.

The issue seems to be the `wrap` parameter (bool) which may have been introduced recently ([doc here](https://neovim.io/doc/user/ui.html#ui-linegrid)).

Adding a `pub wrap: bool` to the `GridLine` struct seems to solve the issue and the build works as expected.

Submitting a pull request.

adnanpri:


vhakulinen:
This would break compatibility with earlier versions of nvim. There is a deeper issue with gnvim's rpc client, which is that it doesn't respect [this](https://github.com/neovim/neovim/blob/5e3cf9fb4bc7f236c858f609ca83c57fb1779ab0/runtime/doc/ui.txt#L115-L117) part of the API contract:

>Future versions of Nvim may add new update kinds and may append new parameters
to existing update kinds. Clients must be prepared to ignore such extensions,
for forward-compatibility.

To fix this, gnvim's msgpack decoding needs to ignore extra values that it is not expecting to get. I'm not sure how this can be done since I haven't had a look at it.

I'm not ready to accept this pull request. We need to either
* Wait for new nvim release and then pump gnvim's nvim version requirement
* Fix the msgpack decoding

As a side note, most of the rpc types are generated straight from nvim's `--api-info` output. See the `scripts/generate-bindings.sh` script.

